PR #9, opened 2017-01-18
Added multispectral example dataset ds117 and ds210

suyashdb:
Added example dataset as per to be proposed multispectral BIDS std.


suyashdb:
json files associated with multispectral data examples are blank.

PR #18, opened 2017-07-07
include EEG example dataset

robertoostenveld:
This is an example dataset prepared by @arnodelorme, following discussion with @CPernet and me. There is still work to be done on this example to align it better with the EEG-BIDS proposal and the other BIDS standards.


chrisfilo:
Thanks for putting this together. I can either merge this into the `eeg` branch or keep it open until EEG-BIDS is ready. What would you prefer?

chrisfilo:
BTW event files should be `_events.tsv` not `_event.tsv` nor `_event.txt`.

robertoostenveld:
I prefer to pull it into a branch rather than keeping it open. I think it is (conceptually) easier for others to pull the branch from INCF than from my repo. And I do want/hope that others contribute.

I suggest renaming the EEG branch to BEP006. That gives a direction for handling future BEPs on the github repository. Could you also make a BEP00x branch for MEG (I don't know the number)?

chrisfilo:
Ok I'm merging this into bep006_eeg. I'll also give you write permission to the repo, but please use PRs to put in changes (it's easier to track progress this way).

I also created bep008_meg.

PR #35, opened 2017-11-03
update to iEEG example with channels versus electrodes fixed

dorahermes:
updates to iEEG  branch example


ieeg_visual/sub-01/ses-01/ieeg/sub-01_ses-01_acq-projected_electrodes.json
```diff
@@ -2,8 +2,9 @@
     "iEEGCoordinateSystem": "mri",
     "iEEGCoordinateUnits": "mm",
     "iEEGCoordinateProcessingDescripton": "surface_projection",
-    "AssociatedImage": "/anat/sub-01_T1w_pial.R.surf.gii",
+    "IntendedFor": "/anat/sub-01_T1w_pial.R.surf.gii",
```
chrisfilo: This `.gii` file should be probably moved to `/derivatives/freesurfer/sub-01/anat/`. `IntendedFor` should also probably be a list pointing to both hemispheres.

choldgraf: our thinking here was that the cortical surface is a pretty crucial thing to have in ecog, moreso than it is in fmri, so we were going to keep it in `anat` rather than have it in derivatives. That said if this is going to be a big convention breaker w/ the other BIDS specs, I dunno...

what do you think @dorahermes?

chrisfilo: What is the cost of keeping it in a different folder (for consistency sake)?

choldgraf: to me I think the biggest cost is extra cognitive load. Right now the BIDS-iEEG spec doesn't mention the `derivatives` folder at all (nor does the poster), so it's another thing we'd have to ask people to wrap their heads around. I'm actually still not super clear on the kinds of things that go in the derivatives folder myself, so it could be that I'm just not conceptualizing this properly.

chrisfilo: I see, but on the other side the derivatives spec is the only place where naming convention for surface files (which is used here) is described. 

Cognitive load aside `IntendedFor` is defined as: _This can be an MRI/CT or operative photo. If only a surface rendering is available, this should point to the surface rendering._ This dataset includes T1w images so shouldn't you point to those?

choldgraf: You make good points here - about to board a flight so I will think about it there. Also curious what @dorahermes thinks.

dorahermes: The current derivatives draft seems to have lost the surface (.gii) specification... For now, we'll keep the surface in the /anat/ folder, as it's also very basic for ECoG and not always derived from the freesurfer output (it often is, but this is not necessary). Wrt cognitive load, the /anat/ folder is the first place for people to look for it. I did correct the IntendedFor image in the example set.

I saved the images in a Matlab array (rather than writing them out as .png/.bmp) and pointed to the array index in the _events.tsv file. I corrected this to be a relative path.

chrisfilo: We can discuss the surface issue at length on the spec or mailing list.

As for the stimuli I would recommend using more common image file formats (.png/.jpg) since they are much more reusable. This is especially important for the example dataset that will be emulated by future users.

robertoostenveld: A model of the cortical sheet (e.g. in a gifti file)  is also required for (some) MEG source reconstruction analysis methods (minimum norm estimation). I think the way it is dealt with in iEEG should be consistent with the way it is dealt with for MEG.

dorahermes:
I corrected the stimuli and made some further updates to align the ieeg example with the extension proposal, can we merge?

ieeg_visual/sub-01/ses-01/ieeg/sub-01_ses-01_task-visual_run-01_ieeg.json
```diff
@@ -1,38 +1,30 @@
 {
     "TaskName": "visual",
-    "SamplingFrequency": 3051.76,
     "Manufacturer": "Tucker Davis Technologies",
-    "ManufacturersModelName": "unknown",
+    "ManufacturersModelName": "",
```
chrisfilo: Wouldn't it make more sense to remove this and similar keys if the value is empty?

chrisfilo:
LGTM - please merge at will.

PR #36, opened 2017-11-10
FIX - fixed events.tsv, which had an empty first header line element …

robertoostenveld:
…and an incorrect first column containing event number (should be onset)

bids-validator returned First column of the events file must be named 'onset' (code: 20 - EVENTS_COLUMN_ONSET)

The number of columns was 10, whereas there are only 9 header labels. The header line started with a tab, suggesting that the first element was removed. Since the first column only contained event number (index), I removed it like this

```
for file in `find . -type f -name \*events.tsv` ; do cut -f 2-20 $file > out && mv out $file ; done
```


robertoostenveld:
following the fix, the bids-validator only reports "This file is too small to contain the minimal NIfTI header", which is expected.

chrisfilo:
Thanks!

PR #41, opened 2017-11-14
added hcp_example_bids dataset

robertoostenveld:
with truncated data files. See issue #40

PR #42, opened 2017-11-14
added pet_phno and hcp_example_bids directories

robertoostenveld:
... from the google drive, with empty raw data files.

This replaces and extends my previous PR #41.


chrisfilo:
Much appreciated!

PR #47, opened 2018-02-02
 ENH: Copy physio data for DS210 from OpenFMRI 

effigies:
Also `chmod -x` a lot of ds210 files.


chrisfilo:
LGTM - but how large is it?

effigies:
About 7.5MB. Maybe a bit much...

PR #48, opened 2018-02-02
ENH: Add synthetic dataset

### LINKED ISSUE #46: Better coverage of all variable types
Right now there are several variable types in the BIDS spec that don't occur in any of the example datasets:

* _stim.tsv.gz
* _confounds.tsv.gz
* _beh.json

Also, the existing `*_physio.tsv.gz` files (which only occur in one dataset) are empty. I get the rationale for not including the full nifti images, but for testing purposes, we should probably include the content of all TSVs.

What would be ideal for my purposes (i.e., for testing pybids properly) would be to have a single dataset that has *all* types of variables represented (i.e., events, physio, stim, confounds, beh, scans, sessions, and participants). The data contained could be nonsensical (though it should be valid). @effigies, if I'm holding you up, can you put something like this together?

effigies:
Closes #46.

Layout:

* Base data set
  * 5 subjects
  * 2 sessions (* 5 = 10 sessions)
  * 1 T1w anatomical (* 10 = 10 anatomicals)
  * 3 functional runs (* 10 = 30 functionals)
    * 2 of one task, 1 of resting state

* Derivatives
  * 2 target spaces (`T1w`, `MNI152NLin2009cAsym`)
  * 1 preprocessed functional (30 * 2 = 60 preprocs)
  * 1 brainmask (30 * 2 = 60 brainmasks)

NIFTI images are all truncated headers (352 bytes, gzipped to ~75 bytes). `nib-ls` outputs as follows:

```
bold.nii.gz      float64 [ 64,  64,  64,  64] 2.00x2.00x2.00x2.50
brainmask.nii.gz float64 [ 64,  64,  64]      2.00x2.00x2.00
T1w.nii.gz       float64 [256, 256, 256]      1.00x1.00x1.00
```

(I can make the dtypes more realistic, if that's an issue.)

#### Physiological data

Physiological data are respiration and cardiac signals, generated as sin waves of 0.5Hz and 1Hz, respectively, and sampled at 10Hz, each file having a unique pair of onset phases.

#### Stimuli

Randomly generated Gaussian noise, sampled at 2Hz, in two columns (`stimA`, `stimB`).

#### Confounds

Confounds are taken to be the first 65 lines (header line + one line per functional volume) of some randomly selected confounds generated in previously run FMRIPREP runs.

#### Note

Skipping behavioral data for now.


effigies:
While this PR was generally intended to fill out sources of information for the `pybids.analysis` module, if it seems useful to have other files/metadata in there, please chime in. It might not make sense to make this dataset all things to all people, but it could probably stand some fleshing out that will make it more useful for testing for pretty little cost.

chrisfilo:
Could you make sure this passes the BIDS validator?

tyarkoni:
The `synthetic` dataset looks great so far, I'm bundling it with pybids. The only thing missing right now that's fairly important from my perspective are `scans.tsv`, `sessions.tsv`, and `participants.tsv` files. I can live without `_beh.json` files, as those are currently used by almost no one (and are probably not going to be included in models very often anyway). But the `.tsv` files are pretty essential for both testing and analysis purposes.

effigies:
Okay, this still doesn't have `_beh.json`, but it does not have:

* `participants.tsv`
* `sessions.tsv`
* `scans.tsv`
* `events.tsv`

Validator now only complains about missing `SliceTiming`.

effigies:
If you've been using this locally, re-fetch. I did a force-push to remove old `.nii.gz` files from the history.

chrisfilo:
This should just be a warning. Could be ignored in this case.

effigies:
Yes. Sorry, should have clarified it's a warning. I dealt with all the errors and relevant-seeming warnings.

sappelhoff:
@chrisfilo, @effigies Why is this not being merged?

PR #60, opened 2018-03-13
fix missing column in participants.tsv

sappelhoff:
I don't really know whether the authors would call it sex, gender, or something similar ... but this column header was missing from the file and preventing it from being properly rendered/readable.


chrisfilo:
God catch!

PR #66, opened 2018-03-21
ds000117 EEGonly added

CPernet:
This is the full data set for EEG only using the latest BIDS spec


chrisfilo:
It seem this PR is mixing commits from multiple sources. 

Additionally `ds000117-EEGonly` datasets seems to use `subXXX` instead of `sub-XXX`. Stimuli files should go into `/stimuli` (lower caps).

chrisfilo:
I'm still seeing a lot of changes that seems unrelated (for example deletion of a lot of files from `7t_trt`). Not sure what happened here. Maybe it would be best to move the changes manually and start over from `bep006_eeg`.

robertoostenveld:
this could be closed if #75 gets merged

robertoostenveld:
The eek-only version has been merged

PR #73, opened 2018-04-13
empty stimuli.png files and added dataset_description.json

dorahermes:
Added dataset_description for the visual example, updated funding in the dataset description for the motor example, and emptied the /stimuli/*.png files in the ieeg_visual example to save space. @choldgraf


choldgraf:
nice! thanks :-)

PR #78, opened 2018-04-20
updates from validator and surfaces to derivatives

dorahermes:
This should address most of the validator issues, @choldgraf can you check? Thanks!!


choldgraf:
woot! let's merge these in, and then I'll run the validator on it to do spot-checks

Are you creating these files automatically, or will it be OK for me to update small parts in a PR? (e.g. if I change something will it be over-written if you then re-generate this BIDS dataset via a script or something like this?)

dorahermes:
I am creating these files automatically... (which does mean that I can update them all rather quickly though, especially handy for the motor dataset)

choldgraf:
ok cool - will keep that in mind!

PR #82, opened 2018-05-15
Added example dataset which includes nested derivatives for ICA and s…

Andesha:
Added example dataset which includes nested derivatives for ICA and segmentation, as well as data property annotation.


Andesha:
It looks as though the BEP006 decisions have not yet been finalized and pushed into the master bids validator yet. As such, I believe that this data set should have been pushed to the BEP006 branch instead.

PR #83, opened 2018-05-15
Added dataset which includes nested derivatives for ICA and segmentat…

Andesha:
Added dataset which includes nested derivatives for ICA and segmentation, as well as data property annotation.


Andesha:
I believe this check will always fail as BEP006 decisions have not yet been finalized and put into the validator.

How should I proceed if I would like this dataset to be shared?

robertoostenveld:
The EEG BEP006 is almost done, but discussion on the EEG derivatives standard BEP021 has just started. Hence I think it would be better to split the PR in two: standard EEG and derivative, and possibly postpone the one with the derivative data until BEP021 takes shape. Right now I am not yet sure how the derivative extensions are going to look. 

You are right that the Travis CI build cannot succeed yet, since it uses the master version of bids-validator, which is unaware of the upcoming EEG extension. So that is not something to worry about. But the sequence will be: get BEP006 merged in the specification, get EEG fully supported in the validator, merge the BEP006 examples branch. 

If you go to http://github.com/robertoostenveld/bids-validator and check out the eeg-validator branch, you can already check the standard EEG part of your example dataset. The validator is not yet 100% correct, so please contribute to squashing the last bugs.

robertoostenveld:
Please see here https://docs.google.com/document/d/1PmcVs7vg7Th-cGC-UrX8rAhKUHIzOI-uIOh69_mvdlw/edit#heading=h.mqkmyp254xh6 for BEP021 - Common EEG Derivatives

robertoostenveld:
I browsed your https://github.com/Andesha/BIDS-examples/tree/eeg_face13_lossless/eeg_face13_lossless and identified multiple issues. 

The data should be at the top-level, not in the "bids" directory. The MATLAB scripts should go in the code directory.  All documentation should go in the standard BIDS top-level documentation files, see section 8.1 in the main BIDS spec. The directory "source" should be called "sourcedata". Please convert the raw EEG data in EDF or in BrainVision format, since (as it looks now) those will most likely be the initially allowed formats for non-processed data. I see multiple files which are not supported according to BIDS EEG, like chan_info.json and time_info.json.

jadesjardins:
Hi Robert, Thanks for browsing the dataset and giving feedback. We will make the corrections you mentioned and make a new pull request. We were also hoping to include some currently unsupported files as examples for the derivatives proposal. These files would be examples of ICA arrays (weights and sphere), as well as data property annotations (chan_info.json, comp_info.json and time_info.json... structures that document extendible classification of channels [e.g. uncorrelated to neighbours, linked/bridged, unusually large voltages, etc] , classifications of components [e.g. blinks, ECG, EMG, occipital alpha, etc], and classification of time points [out-of-task, in-task, non-stationary noise, etc]). Would you prefer that we leave these files out of our example before discussing these topics in the derivatives proposal?

jadesjardins:
Sorry, I just reread the earlier comment, we will separate the PR into standard EEG and then wait to include a derivative set.

dorahermes:
just tagging @choldgraf for the validator

robertoostenveld:
The repo contains the bids-validator command line utility in the bin directory. So assuming installed in $HOME this would work: ~/bids-validator/bin/bids-validator <dataset>

sappelhoff:
great thanks. Where should I report issues with the EEG-validator? On your fork, issues seem to be disabled

robertoostenveld:
I have just enabled issues on my copy of bids-validator <https://github.com/robertoostenveld/bids-validator>. That means you can also send PRs to me directly.

sappelhoff:
@Andesha, @jadesjardins can this PR be closed? We need to clean up the stale EEG PRs a bit to create some clarity about what could potentially be merged soon and what not :-)

PR #86, opened 2018-05-26
Fix embla

### LINKED ISSUE #81: EMBLA sleep example should be reviewed
Hi @ChristophePhillips and @CPernet, I have cleaned up and converted the EMBLA example in line with the latest version of the EEG draft at http://bit.ly_bids_eeg. Could you review the organization of the example dataset?

sappelhoff:
fixes #81 

besides these fixes the reformatting of the embla dataset looks compliant to me.

However, if we decide to allow for types "GRD" and "REF" in the channels.tsv, then we'd have to adjust that again (i.e., the current type of REF in _channels.tsv is EEG ... but should be changed to REF, if we should allow that type)

### DIFF
--- a/eeg_sleep/sub-01/eeg/sub-01_task-sleep_channels.tsv
+++ b/eeg_sleep/sub-01/eeg/sub-01_task-sleep_channels.tsv
@@ -5,7 +5,7 @@ Chin	EMG	uV
 Cz	EEG	uV
 E1	EEG	uV
 E2	EEG	uV
-EKG	EEG	uV
+EKG	ECG	uV
 F3	EEG	uV
 F4	EEG	uV
 Fz	EEG	uV
--- a/eeg_sleep/sub-01/eeg/sub-01_task-sleep_eeg.json
+++ b/eeg_sleep/sub-01/eeg/sub-01_task-sleep_eeg.json
@@ -6,5 +6,6 @@
   "EMGChannelCount": 1,
   "ECGChannelCount": 1,
   "MiscChannelCount": 3,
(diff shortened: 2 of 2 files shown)


robertoostenveld:
I don't think that GND and REF as channel type make sense, since those would (normally) not even be present as channels in the dataset (since only used in the analog part of the amplifier and not digitized). But in this case the source data has a REF.ebm file, just like e.g. a C3.ebm file. I don't have the actual files, only the 0-byte ones, and don't know anything about the EEG system. That is why I kept it in as channel. Perhaps @ChristophePhillips can clarify.

sappelhoff:
@robertoostenveld, how would you suggest that we specify ground and reference? In the `*_eeg.json`?
```
...
[EEGReference]
[EEGGround]
...
```
?

robertoostenveld:
In the json file you can specify the free-form description of EEGReference. I think it would be good to also include EEGGround in the json and proposed it in the BIDS-EEG gdoc I have added that to the draft.

I have also added following text "Note that for proper documentation of EEG recording metadata it is important to understand the difference between electrode and channel: An EEG electrode is attached to the skin, whereas an channel is the combination of the analog differential amplifier  and analog-to-digital converter that result in a potential (voltage) difference that is stored in the EEG dataset. Although the  “reference” and “ground” are often referred to as channels, they are in most common  EEG systems not amplified and recorded by themselves, and therefore should not be represented as channels but as electrodes. The type of referencing and (optionally) the location of the reference and electrode and  the location of the ground electrode should be specified. “

I have also elaborated on columns for electrodes.tsv and added REF and GND to the example there.

PR #84, opened 2018-05-24
Bep006 eeg

CPernet:
this fork contains EEG data only (not sure if easier to merge) all updated for the last specification

PR #87, opened 2018-05-26
add the Cuban EEG example dataset

robertoostenveld:
see PR #84


robertoostenveld:
sorry, this should not go into master

PR #92, opened 2018-05-26
Add dataset matchingpennies

sappelhoff:
A new dataset for testing EEG BIDS. Sourcedata in `xdf` format, converted to brainvision. See on [OSF](https://osf.io/cj2dr/)


robertoostenveld:
I presume you tested this one? I would prefer not to add more failing example datasets, but rather squash out the bugs in the existing examples...

sappelhoff:
Yes, the only "error" is the one of inheritance, see [issue on your fork](https://github.com/robertoostenveld/bids-validator/issues/1)

PR #93, opened 2018-05-27
TYPO: eeg_matchingpennies dataset, fix link in README

sappelhoff:

PR #94, opened 2018-05-28
Fix eeg validator AND typofix

sappelhoff:
This should make #93 pass the test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ node_js:
   - "6.11.0"
 
 install:
-  - npm install -g https://github.com/INCF/bids-validator.git#eeg-validator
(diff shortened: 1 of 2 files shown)


sappelhoff:
okay, there is an issue with [eeg_rishikesh](https://github.com/INCF/BIDS-examples/tree/bep006_eeg/eeg_rishikesh) ... but at least the validator does its work :-)

.travis.yml
```diff
@@ -2,7 +2,7 @@ node_js:
   - "6.11.0"
 
 install:
-  - npm install -g https://github.com/INCF/bids-validator.git#eeg-validator
+  - npm install -g https://github.com/robertoostenveld/bids-validator.git#eeg-validator
```
robertoostenveld: ah, that was why it still did not work! thanks

PR #96, opened 2018-06-13
[WIP] updated bep010 ieeg examples

robertoostenveld:
I tested them against the bids-validator and compared them to the draft specification in the google doc.


robertoostenveld:
@choldgraf and @dorahermes can you please check this?

choldgraf:
seems good to me - I take it you mean the version of the bids-validator in the PR that you've opened to my branch, yah?

robertoostenveld:
yes, it was that version of the validator.

dorahermes:
Thank you, these are super useful updates! We are moving in 1 week, so I do not have time for a thorough check now. 

I auto-generated all metadata files with scripts from the raw data and would be able to update them all quickly when we finalize our thoughts on the surfaces and coordinate systems naming when @chrisfilo is back from HBM.

choldgraf:
cool (as this PR is still labeled WIP I'm not going to merge it, but @robertoostenveld let us know if it's ready to go, we'll need to spot-check subsequent changes to the specification in the future anyway)

sappelhoff:
what's the status of this PR? Does it also fix #100 ?

robertoostenveld:
How about the "T1w" coordinate system? It ain't in the list in the main BIDS draft and it is not clear from T1w what the origin or the direction of the axes is. 

"Other" is an option though, which then requires iEEGCoordinateSystemDescription to be used. For example iEEGCoordinateSystemDescription = "This is in relation to the original anatomical MRI of the participant (not shared), which is expressed in GE scanner device coordinates." I still would not know what GE scanner device coordinates would be, but presumably a GE expert could tell me.

This would imply that the files need to be renamed from 
   sub-ca_ses-01_space-T1w_coordsystem.json
 to 
   sub-ca_ses-01_space-Other_coordsystem.json

robertoostenveld:
iEEGReference and SamplingFrequency are missing from (some of) the ieeg.json files.

sappelhoff:
There are still some remaining issues, see: https://travis-ci.org/INCF/BIDS-examples/builds/427561686

robertoostenveld:
The problem that I still have now is that the xxx_space-2Dimage_coordsysystem.json file is not allowed. I think this requires some regexp work, but that seems to be on hold for now.

PR #101, opened 2018-06-19
Fix rishikesh

sappelhoff:


sappelhoff:
The validator warnings for this dataset all related to the - not yet implemented - inheritance principle.

I mostly deleted duplicate files and added required fields.

PR #109, opened 2018-07-24
Fix rishikesh

sappelhoff:
yet another fix of the EEG examples that became necessary after updating the EEG-validator


sappelhoff:
closed in favor of #110

PR #110, opened 2018-08-10
[MRG] Fix eeg examples

sappelhoff:
This is rather big because I am adjusting the complete directory of `ds000117_eeg_only` to follow the conventions of our remaining examples.

Also some other updates for the examples in accordance with our current spec.


sappelhoff:
@robertoostenveld perhaps you could do a review and merge if you agree?

robertoostenveld:
it looks good as far as I can see. I only noticed this section as unusual

```
"HardwareFilters": {
        "Lowpass butterworth filter": {
             "Cutoff (Hz)": 350
         }
```
as it would mean that the name of the json field has a space and brackets. I don't think that the units should be part of the key, as it is specified that "unless otherwise specified, units are SI". 

I don't know where it is specified how HardwareFilters can be specified. This way also allows for the specification of a filter order, etc, so is fine with me. But the previous specification was also OK with me.

I recommend you remove the "(Hz)"

sappelhoff:
> I don't know where it is specified how HardwareFilters can be specified.

See here:

https://github.com/robertoostenveld/bids-validator/blob/ee265af723cb38df6c1da808cf0a27e4bc87410a/validators/schemas/eeg.json#L29-L34

In our spec we say:

`HardwareFilters: OPTIONAL - List of hardware (amplifier) filters applied or ideally  key:value pairs of pre-applied filters and their parameter values.`

The example in our spec:

`"HardwareFilters": "ADC's decimation filter (hardware bandwidth limit), 5th order sinc response with a -3 dB point at 480Hz"`

We should update that example to conform to our code in the validator. It would then look like this:

```
"HardwareFilters": {
        "ADC's decimation filter (hardware bandwidth limit)": {
             "-3dB cutoff point (Hz)": 480,
             "Filter order sinc response": 5
         }
```

Personally, I think keeping the (Hz) explicit is helpful for human reading, which is our priority as this field is not machine-parsed at this moment.

In any case, formatting the hardware filters as "objects" is more explicit / readable than a long string, I think :-)

sappelhoff:
@chrisfilo @robertoostenveld 
if there are no objections, please merge while we are wrapping up

chrisfilo:
@robertoostenveld I'm leaving it up to you.

robertoostenveld:
Hi @sappelhoff, sorry for me being slow in responding. 

FYI - elsewhere in BIDS specs the specification "list" is used as a list with square brackets. See e.g. dataset_description.json in http://bids.neuroimaging.io/bids_spec.pdf. But then again, SoftwareFilters for MEG is a curly-bracketed list. 

Somehow "HardwareFilters" did not make it into the MEG specification, which I consider an omission that should be fixed.  I think that (as with common fields for imaging data), we should also make a list of fields that is common for MEG/EEG/iEEG.

Right now I am working on https://github.com/robertoostenveld/fieldtrip/blob/bids/data2bids.m where etc similarities (and differences) are also relevant.

sappelhoff:
no worries @robertoostenveld. The validator currently asks for either curly brackets (i.e., dictionaries aka json objects) or a string "n/a":

https://github.com/INCF/bids-validator/blob/307b4b2e9aa6381c47e0b2f935b0bb081376f18c/validators/schemas/eeg.json#L29-L40

If we want to specify filters in an array format (i.e., using `[ ]`), we need to change that in the validator. However, how would you propose that filters are being specified as arrarys? Can you give an example?

I think that the current way of specifying hardware and software filters as nested dictionaries allows for a versatile, comprehensive, and clear specification (human readable) ... this way allows for multiple filters with multiple parameters each

--> such as whether the filter is lowpass, highpass, bandpass,, ... and the filter type (butterworth, rc, ...), cutoff frequency (in Hz ... and also type: half-power, half-amplitude), filter roll-off, ...


> Somehow "HardwareFilters" did not make it into the MEG specification, which I consider an omission that should be fixed. I think that (as with common fields for imaging data), we should also make a list of fields that is common for MEG/EEG/iEEG.

I completely agree

CPernet:
I'm just not sure of hardware filters required
- people might not know about them, specially older dataset we want to push (i know it's the case for me)
- given that we apply filters on the data anyway (software) it makes little difference (it does to some point of course)

i have no strong feelings about it though, just stating my opinion

dorahermes:
Summarizing my recollection of the history for iEEG:
- We initially had both hardware and software filters as 'required'.
- We had several contributors request that hardware filters not be required, so this was changed to recommended. 
- Since software filters can be n/a in case there were none, there were no comments on this.

robertoostenveld:
I do not know of any professional system that does not implement a hardware filter prior to digitization (since it increases noise). I do know of researchers that somehow think that - if they configure the filter in the acquisition software - it is therefore a software filter (although the software often sends the parameters to the hardware which does the filtering). But sometimes it is difficult to distinguish between hardware, firmware and software filters.  

Having something required (with n/a as possible) value is the safest option, we could always be more lenient in the future, as that would not invalidate existing datasets. 

I value consistency with MEG, which currently does not have the HardwareFilter option. I propose that we therefore add HardwareFilter to MEG as OPTIONAL (= action 1), and that we keep SoftwareFilter as REQUIRED for all three.

We should also document it consistently  (= action 2). Right now `HardwareFilters` for iEEG is too short, i.e. it does not explain that it can be a list, or that n/a is a valid option.

robertoostenveld:
> see above action 1 and action 2

I just did both in https://github.com/sappelhoff/bids-specification/pull/6

Issue #100, opened 2018-06-15
TODO: BIDS-iEEG filtered speech updates


choldgraf:
* Updated `iEEGCoordinateUnits` from "indices" to "pixels"

(I'll update this as more stuff comes up)

PR #125, opened 2018-10-02
renamed coordsystem files to make the file names compliant

robertoostenveld:
with the validator.


robertoostenveld:
see also #100

choldgraf:
I don't believe that the travis errors (all BIDS validation errors) are related to the file-name changes here, so I'm gonna merge!

Issue #132, opened 2018-11-20
To Do for merging bep006 examples


sappelhoff:
Feel free to add missing points.

see https://github.com/bids-standard/bids-specification/issues/107 for the roadmap 

cc: @robertoostenveld @CPernet @arnodelorme @gllmflndn @ChristophePhillips

- [x] finish updating the bids-validator https://github.com/bids-standard/bids-validator/issues/592
- [x] re-format `eeg_matchingpennies` according to bids-validator `master` as updated in step 1.
    - [x] upload reformatted data to OSF: https://osf.io/cj2dr/
- [x] remove `eeg_sleep` example unless @robertoostenveld disagrees
- [x] update eeg_rishikesh with 1.0 version of data on zenodo: https://zenodo.org/record/1490922
- [x] wait for @arnodelorme to release version 1.1 of the rishikesh data on zenodo, hopefully up to date according to bids-validator `master` as updated in step 1.
    - [x] push zenodo rishikesh version as 0kb to the `bep006` branch, replacing the current `eeg_rishikesh`
- [x] where did the "cuba resting state" dataset by @CPernet go?
- [x] merge cuba dataset, and add information about it on BEP006 google doc, as well as WIKI in this repo
- [x] re-format `eeg_resting-fmri` according to bids-validator `master` as updated in step 1.
    - [x] upload reformatted data to OSF: https://osf.io/94c5t/
- [x] re-format `eeg_ds00017` according to bids-validator `master` as updated in step 1.
    -  upload reformatted data to ~openneuro: https://openneuro.org/datasets/ds000117/versions/1.0.2~ ... we can do that only once BEP006 has been *completely* merged
- [x] update .travis so that no `--bep006` flag is used
- [x] ~in resting state fmri data, we need to rename the `.dat` eeg files to `.eeg`~ only in derivatives, so leave as is

sappelhoff:
>  --> so far it passed bids validator -bep006

Yes, but through the updating process of the validator (which always lags a bit behind), we are getting new errors now. See:

> Validating dataset eeg_rest_fmri/
	1: [ERR] Invalid JSON file. The file is not formatted according the schema. (code: 55 - JSON_SCHEMA_VALIDATION_ERROR)
		./task-rest_eeg.json
			Evidence:  should have required property 'SoftwareFilters'

So it should be fine by setting `"SoftwareFilters": "n/a"`.

[...]

PR #134, opened 2018-12-03
[WIP] bring matchingpennies up to date with spec

sappelhoff:
related to #132


sappelhoff:
This will pass once https://github.com/bids-standard/bids-validator/pull/663 is merged.

Except for the EMPTY FILE ERROR, which we still have for all example data.

PR #141, opened 2018-12-06
cb data

CPernet:
added the cuban data - reminder:acq_time error in scans.tsv ??


sappelhoff:
Here's the answer to the issue: https://neurostars.org/t/scans-tsv-acq-time-error/2982/2

sappelhoff:
also, you should add a `participants.json` at the root, so that age, sex, etc. from the `participants.tsv` are described. See e.g., https://github.com/bids-standard/bids-examples/blob/bep006_eeg/eeg_matchingpennies/participants.json

sappelhoff:
looks good, and it passes - so in principle ready to merge. 

Just one question: in the participants.json you say

>         "Units": "years and months"

```
participant_id	age	sex
sub-cbm001	20.94	M
sub-cbm002	46.57	F
sub-cbm003	24.16	M
sub-cbm004	29.66	M
```

Looking at the data, I can't really interpret what you mean ... if it is years and month, I would expect something like 11.11, for a subject that is 11 years and 11 months old ... but in your data we have something like 20.94 ... so it is probably not 20 years and 94 months.

Do you mean 20.94 years with that? i.e., 7455 *days* old? (note that `7455/356 = 20.94`)

CPernet:
it's not my data ... but yes I guess the value is just years then?

sappelhoff:
> it's not my data ... but yes I guess the value is just years then?

yes, let's go with years. I can merge once that change is incorporated

sappelhoff:
thanks @CPernet

PR #143, opened 2018-12-11
replace encoding error with na

sappelhoff:
I am replacing the description `�V` with `n/a`. 

OK with you @CPernet ?


CPernet:
did i not changed that already? sure do

PR #144, opened 2018-12-12
Bep010 ieeg visual and motor examples updated

dorahermes:
Both sets passed the validator. There are only some warnings of custom columns left, which does not seem correct.
@choldgraf


dorahermes:
@choldgraf can we merge this or do you want to update the validator first?

choldgraf:
I'll take a pass at the validator to finish it up again soon, was spending my cycles on the paper and now in holidays mode, but will get back on this after the new year begins

sappelhoff:
I just closed and reopened this to run Travis against the new config and it turns out that the "pass" is deceiving: looking at the details, there are still some issues with `ieeg_motorMiller2007` and `ieeg_visual`.

Do you have time to address these issues @dorahermes ? https://travis-ci.org/bids-standard/bids-examples/builds/500215549?utm_source=github_status&utm_medium=notification

dorahermes:
Yes, will correct them and make a pull request later today.

dorahermes:
I have one error that I can not resolve: we had a non-integer sampling frequency (1525.9 Hz), but the validator wants sampling frequency to be an integer: Evidence: .SamplingFrequency should be integer.

dorahermes:
I addressed all `ieeg_motorMiller2007` and `ieeg_visual` issues

PR #145, opened 2019-01-10
update eeg_rishikesh to its 2.0 version

sappelhoff:
related to #132 

eeg_rishikesh has now been updated to comply with our changes in BEP006 since its initial release.

This has been formatted by @arnodelorme and reviewed by me, so I will merge this PR once the CIs are green


sappelhoff:
It turned out that we missed the "SoftwareFilters" during the review and the validator did not catch it. This is because we used an old version, since the most recent one seemed to "run out of memory", see https://github.com/bids-standard/bids-validator/issues/677

I added the missing fields here ... however in the data on zenodo, these are still missing.

PR #148, opened 2019-02-28
updating ieeg filtered speech and deleting unnecessary files

choldgraf:
This removes some extra files (edf files etc) from the earlier upload that weren't deleted. It also adds a README.


choldgraf:
@sappelhoff arggg sorry I totally didn't see your comment before merging. that's a good catch, there must be a bug in my conversion code. I'll make another PR to fix (either way, good to get this PR in now because the current setup of this example folder had duplicated data files)

PR #151, opened 2019-03-11
Merge Bep010 ieeg examples onto master

### LINKED ISSUE #131: iEEG examples to add
The following is a list of example datasets that would be particularly useful to highlight the iEEG specification for BIDS:

* multiple runs of clinical observation data
* Clinical recordings during no task
* Seizure recordings
  We could maybe add the dataset we use for the Brainstorm SEEG tutorial?
  https://neuroimage.usc.edu/brainstorm/Tutorials/Epileptogenicity
* An example showing how to properly format hardware/software filters:
  
  <details>

  I am linking to an EEG example on how the filter fields could be formatted: 

  https://github.com/bids-standard/bids-starter-kit/wiki/FAQ#how-to-format-hardware-and-software-filter-fields-in-a-json

  in brief, it means defining a dictionary for each filter field. This dictionary then 
  contains a filter-name, which again links to a dictionary ... this dictionary then 
  contains the key value pairs of the filter parameters. 

  I agree that it sounds a bit complicated and nested at first glance, but it allows for a 
  very exhaustive documentation of the filters. For example two different kinds of 
  hardware filters with different sets of parameters

  feedback is welcome, since this is not very fixed in EEG.
  </details>

sappelhoff:
closes #131 
Apart from rebasing on master, this PR brings two changes:

1. bump up travis version: https://github.com/bids-standard/bids-examples/commit/49cc2009e3a23e699fe9224967b00ae017ad0966
2. zero out two large datafiles in `ieeg_visual`: https://github.com/bids-standard/bids-examples/commit/cfac0a4cbd94eab7f299dceb96a03443eab5a364


sappelhoff:
this is relevant to: https://github.com/bids-standard/bids-validator/issues/749

I'd like to merge these examples soon and then make a release of the `bids-examples` to be tested with the `bids-validator`.

One more approving review from @dorahermes or @choldgraf should do the trick :-)

ieeg_epilepsy/sub-01/ses-postimp/ieeg/sub-01_ses-postimp_space-MNI152_electrodes.tsv
```diff
@@ -0,0 +1,125 @@
+name	x	y	z	size	hemisphere	group	type	manufacturer
+v'1	-1.3	-39	31	5	L	v'	depth	DIXI
```
dorahermes: name: Label of the channel, only contains letters and numbers.

ieeg_epilepsy/sub-01/ses-postimp/ieeg/sub-01_ses-postimp_space-MNI152_coordsystem.json
```diff
@@ -0,0 +1,8 @@
+{
+    "iEEGCoordinateSystem": "T1w",
```
dorahermes: Should be one of the listed keywords, is this 'MNI152NLin6Sym' or 'Other'? (other for non-normalized MRI)

ieeg_epilepsy/sub-01/ses-postimp/ieeg/sub-01_ses-postimp_space-head_coordsystem.json
```diff
@@ -0,0 +1,8 @@
+{
+    "iEEGCoordinateSystem": "T1w",
```
dorahermes: iEEGCoordinateSystem should be 'Other'

ieeg_motorMiller2007/sub-bp/ses-01/ieeg/sub-bp_ses-01_space-T1w_electrodes.tsv
```diff
@@ -0,0 +1,48 @@
+name	x	y	z	size	type	manufacturer
+1	-38.2367221940641	42.990911138808	32.1161296015272	2.3	surface	AdTech
```
dorahermes: I should correct the size to surface area (currently diameter)

ieeg_epilepsy/sub-01/ses-postimp/ieeg/sub-01_ses-postimp_space-MNI152_electrodes.tsv
sappelhoff: mmh this might be complicated if in the raw data the channels have other symbols in their label ... do you have a suggestion for a fix that would not involve reading in all data and renaming the labels?

ieeg_motorMiller2007/sub-bp/ses-01/ieeg/sub-bp_ses-01_space-T1w_electrodes.tsv
sappelhoff: While you are at fixing your own dataset, could you correct the epilepsy one as well to have "Other" instead of "T1w" (as you noted above)? :-)

dorahermes: yes! I should make a pull request to your fork/branch, correct? (will be tomorrow)

sappelhoff: yes, to https://github.com/sappelhoff/bids-examples/tree/bep010_ieeg would be ideal, thanks!

dorahermes: just made a pull request to your fork/ieeg branch!

sappelhoff:
@dorahermes do you approve of merging this PR now?

dorahermes:
The only comment left is that the electrode names in the ieeg_epilepsy dataset contain a non-letter/number symbol: '. I cannot fix this according to the data/header, because they have zero bites and I can't read the channel labels. I could just remove the ', or just leave it since the dataset currently passes the validator? @choldgraf

sappelhoff:
What was the reason to only allow numbers and letters in the first place? We don'T have this restriction in EEG, and apparently we missed this inconsistency between EEG and iEEG :grimacing:

choldgraf:
arggg - what's the state of the validator? I'm guessing it'll pass both of them even if there's a symbol in the channel names, is that right?

sappelhoff:
> I'm guessing it'll pass both of them even if there's a symbol in the channel names, is that right?

yes, that's right :-)

why do you think that symbols in the channel name are a bad idea?

dorahermes:
The initial thought was that it would be more difficult to match electrode names and channels labels in case of symbols.

sappelhoff:
> The initial thought was that it would be more difficult to match electrode names and channels labels in case of symbols.

Do you have an example where it would be more difficult? Just so that I can understand better?

To move further, I see these options:

1. we change the spec to also allow symbols ... this would not be backward compatibility breaking
2. we enforce to not allow symbols. This would mean:
    - asking @ftadel to adjust his dataset so that channel/electrode names do not contain symbols
    - implementing a check in the validator that prevents symbols in channel/electrode names

choldgraf:
if symbols can be in both the channels and in the electrodes names, then I'd be +1 on supporting them since those symbols would be "just another character in a string". 

e.g., If channels and electrodes were named `ch-01` I could see that being easier to parse then `ch01`

choldgraf:
(but, this would require a change to the spec, as @sappelhoff says...so to me the question breaks into "do we change the spec, or do we change the validator")

sappelhoff:
+1 for adding to the spec that symbols are also permissible

dorahermes:
+1 for adding to the spec that symbols are also permissible for the same reasons

ftadel:
+1 for allowing symbols
Many epilepsy centers use a prime (') to indicate whether an electrode is implanted in the left or the right hemisphere (eg. leading to channels A01..A12 and A'01..A'12). Not allowing the single quotes and some other variations (commas, dots, ...) would make the formatting into BIDS more complicated.

sappelhoff:
alright! see bids-standard/bids-specification#182 ... please review :-)

sappelhoff:
just fyi: Unless something happens, I will merge this one once bids-standard/bids-specification#182 is merged

choldgraf:
@sappelhoff I think that per the new governance document, we need to leave this open for a while longer, no?

choldgraf:
I think @dorahermes should "un-request changes" before we merge, just to respect the community process...

sappelhoff:
> I think @dorahermes should "un-request changes" before we merge, just to respect the community process...

+1

choldgraf:
🎉

PR #155, opened 2019-04-19
New ECOG epilepsy example

ftadel:


dorahermes:
thanks! lgtm

ieeg_epilepsy_ecog/sub-ecog01/ses-postimp/ieeg/sub-ecog01_ses-postimp_task-seizure_run-01_ieeg.vhdr
```diff
@@ -0,0 +1,119 @@
...
+
+[Binary Infos]
+BinaryFormat=IEEE_FLOAT_32
+
+[Channel Infos]
+; Each entry: Ch<Channel number>=<Name>,<Reference channel name>,
+; <Resolution in "Unit">,<Unit>, Future extensions..
+; Fields are delimited by commas, some fields might be omitted (empty).
+; Commas in channel names are coded as "\1".
+Ch1=G_A1,,1,�V
```
sappelhoff: could it be that there is some encoding error going on here?

ftadel: Fixed

ftadel:
Please do NOT merge now: the neurologists in Freiburg just realized that the implantation scheme is wrong, the order of the contacts in the ECoG grid is incorrect...

sappelhoff:
Is this ready for review now @ftadel ?

ftadel:
Yes

ieeg_epilepsy_ecog/sub-ecog01/ses-postimp/ieeg/sub-ecog01_ses-postimp_task-seizure_run-01_channels.tsv
```diff
@@ -0,0 +1,98 @@
+name	type	units	low_cutoff	high_cutoff	sampling_frequency	reference	group	status
+G_A1	ECOG	uV	0.16	344	1024	intracranial	G	good
```
sappelhoff: `uV` is currently not BIDS ... but there is no validator check in place, so this passes.

We accept `µV` or `microV`, see the [Appendix V](https://bids-specification.readthedocs.io/en/stable/99-appendices/05-units.html)

Note that there is currently an issue discussing to allow ASCII units like `uV`: https://github.com/bids-standard/bids-specification/pull/151 Feel free to chime in with your opinion on these matters! :-)

ftadel: I understand the idea, we don't want to have too many redundant examples, a few complete examples will do a better standardization job and will be easier to update with the following versions of the specification.

This example is the only one that illustrates a patient with implantation schemes and anatomical images at the three times of the surgery (pre-implantation, post-implantation, post-surgery). As the specification is not providing any guidelines for how to name the different sessions and how to distribute the various files (where to put the implantation schemes?), this could be used as a typical recommended structure instead. 
These anatomical informations do not seem to be of interest for the other ECoG examples that have been posted here, but are mandatory for any accurate modeling or visualization.

It also includes both SEEG and ECOG, which can be a good way to test that both can be handled simultaneously (useful to validate analysis programs such as FieldTrip, SPM or Brainstorm).

If you want to keep less examples, we could think about removing the previous example I posted (ieeg_epilepsy), which is less complete (only SEEG, no post-operative images, no implantation schemes).  That one is interesting as an epilepsy case, but not necessary as an ontology example. 

> We accept µV or microV, see the Appendix V

Thanks for spotting this.
It's probably better to accept less options for units formatting. If we accept too many variations, the programs processing automatically BIDS datasets might miss the interpretation of some of the extra options. This is "useless" freedom (who cares if the file has to use microV instead of uV?) might make development/maintenance/debugging more complicated.
But the validator should enforce this properly then.

ftadel:
I understand the idea, we don't want to have too many redundant examples, a few complete examples will do a better standardization job and will be easier to update with the following versions of the specification.

This example is the only one that illustrates a patient with implantation schemes and anatomical images at the three times of the surgery (pre-implantation, post-implantation, post-surgery). As the specification is not providing any guidelines for how to name the different sessions and how to distribute the various files (where to put the implantation schemes?), this could be used as a typical recommended structure instead.
These anatomical informations do not seem to be of interest for the other ECoG examples that have been posted here, but are mandatory for any accurate modeling or visualization.

It also includes both SEEG and ECOG, which can be a good way to test that both can be handled simultaneously (useful to validate analysis programs such as FieldTrip, SPM or Brainstorm).

If you want to keep less examples, we could think about removing the previous example I posted (ieeg_epilepsy), which is less complete (only SEEG, no post-operative images, no implantation schemes). That one is interesting as an epilepsy case, but not necessary as an ontology example.

sappelhoff:
thanks @ftadel!

PR #157, opened 2019-05-04
Use Travis to block PRs with large files

### LINKED ISSUE #156: Use git hooks to prevent files of size >= 1MB from being committed
I don't know how exactly, but this should be possible. See this [general introduction](https://githooks.com/). It would prevent issues such as the one we have to resolve in #119 

In short, each time somebody wants to commit changes to `bids-examples`, a small script would be run to check that all changes are < 1MB in size. If the size requirement is exceeded, the commit is blocked and an appropriate error message is shown.

sappelhoff:
fixes #156 

I am trying to ask Travis to check for file sizes now, blocking everything > 500kb.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ node_js:
   - "8.0.0"
 
 install:
-  - npm install -g bids-validator@1.2.1
+  - npm install -g bids-validator@1.2.3
 
 script:
   # Validating the BIDS datasets using bids-validator
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+language: "node_js"
+
 node_js:
   - "8.0.0"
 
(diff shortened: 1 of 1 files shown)


choldgraf:
can we test if this works by committing a bit file to this PR once and seeing if travis fails? we can use `git commit --amend --no-edit` so it doesn't get committed to the git history

sappelhoff:
I added a `./bigfile` of 1MB size, and Travis caught it, see the log: https://travis-ci.org/bids-standard/bids-examples/builds/528367294?utm_source=github_status&utm_medium=notification

I'll revert that commit now, then git amend, and push -f

choldgraf:
sweet! I'll give this a merge and hopefully it'll work moving forward! Thanks for the addition @sappelhoff

PR #159, opened 2019-05-04
Migrating Documentation from WIKI to README

### LINKED ISSUE #74: sub-folders for modalities
I think it'd be helpful if there were a way to distinguish examples for particular modalities. This could either be something like appending the modality name to each folder (e.g. `ieeg_ds100`), or we could put them in sub-folders (`BIDS-examples/meg/ds100` and `BIDS-examples/fmri/ds101`). There will be many datasets that have multiple modalities, but I think this would be useful to guide newcomers and make it easier for them to find the information they want.

### LINKED ISSUE #154: Add links to download the full iEEG examples
How do we download the full datasets?
- ieeg_visual_multimodal: @irisgroen
- ieeg_visual: @dorahermes
- ieeg_motorMiller2007: @dorahermes 
- ieeg_filtered_speech: @choldgraf
For all datasets: @sappelhoff @robertoostenveld

Now the BIDS-validator is iEEG-compatible: could you upload them on openneuro and add the download links to the README file of this repo?

sappelhoff:
This PR fixes #74, by making the documentation on available datasets more accessible and providing a table that shows (approximately) which modalities are present in each dataset.

Also fixes #154 to some extend, by adding all available links to the full data. The rest will have to be added by the respective dataset-authors later on.

The table is quite large although I tried hard to cut down on complexity. I would be very grateful for a review and some feedback.

Note that I also inserted a comment in the README, that we are interested in new examples **as long as** ... they cover some aspects that are not already covered. 

I don't see a point of hosting many datasets here if they are all the same in terms of their structure. Personally, I would even be up for cutting down on some example datasets we are hosting here - so please let me know about your opinion on this as well.

- [ ] Upon merge, delete Wiki

--- 
Rendered version: https://github.com/sappelhoff/bids-examples/blob/overhaul_docs/README.md

Google Sheet version: https://docs.google.com/spreadsheets/d/1d9KfmaKvTOpFWgi0Kd-tZGgVTsZk_4BHH9prcgFeigw/edit?usp=sharing

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,19 +29,19 @@ corresponding Pull Request.
 
 |  name | description | mri | meg | eeg | ieeg | full data |
 | --- | --- | --- | --- | --- | --- | --- |
-|  7t_trt | field maps, physiological data, quantitative T1 maps, T1w, BOLD | x |  |  |  | https://bit.ly/2H0Z6Qt |
-|  ds000117 | A multi-subject, multi-modal human neuroimaging dataset of 19 subjects on a MEG visual task | x | x | x |  | https://openneuro.org/datasets/ds000117/ |
-|  ds000246 | Auditory dataset used for Brainstorm’s general online tutorial | x | x |  |  | https://openneuro.org/datasets/ds000246/versions/00001 |
-|  ds000247 | Five minutes, eyes-open, resting-state MEG data from 5 subjects. This is a sample from The Open MEG Archive (OMEGA). | x | x |  |  |  |
-|  ds000248 | MNE sample data: Data with visual and auditory stimuli | x | x |  |  | https://bit.ly/2JfNYkf |
-|  ds001 | single task, multiple runs, in-plane T2, events, T1w, BOLD | x |  |  |  | https://bit.ly/2H0Z6Qt |
-|  ds002 | multiple tasks, multiple runs, in-plane T2, events, T1w, BOLD | x |  |  |  | https://bit.ly/2H0Z6Qt |
-|  ds003 | single task, single run, in-plane T2, events, T1w, BOLD | x |  |  |  | https://bit.ly/2H0Z6Qt |
-|  ds005 | single task, multiple runs, in-plane T2, events, T1w, BOLD | x |  |  |  | https://bit.ly/2H0Z6Qt |
-|  ds006 | single task, multiple sessions, multiple runs, in-plane T2, events, T1w, BOLD | x |  |  |  | https://bit.ly/2H0Z6Qt |
-|  ds007 | single task, multiple runs, in-plane T2, events, T1w, BOLD | x |  |  |  | https://bit.ly/2H0Z6Qt |
-|  ds008 | multiple tasks, multiple runs, in-plane T2, events, T1w, BOLD | x |  |  |  | https://bit.ly/2H0Z6Qt |
-|  ds009 | multiple tasks, multiple runs, in-plane T2, events, T1w, BOLD | x |  |  |  | https://bit.ly/2H0Z6Qt |
+|  7t_trt | field maps, physiological data, quantitative T1 maps, T1w, BOLD | mri |  |  |  | https://bit.ly/2H0Z6Qt |
+|  ds000117 | A multi-subject, multi-modal human neuroimaging dataset of 19 subjects on a MEG visual task | mri | meg | eeg |  | https://openneuro.org/datasets/ds000117/ |
+|  ds000246 | Auditory dataset used for Brainstorm’s general online tutorial | mri | meg |  |  | https://openneuro.org/datasets/ds000246/versions/00001 |
+|  ds000247 | Five minutes, eyes-open, resting-state MEG data from 5 subjects. This is a sample from The Open MEG Archive (OMEGA). | mri | meg |  |  |  |
+|  ds000248 | MNE sample data: Data with visual and auditory stimuli | mri | meg |  |  | https://bit.ly/2JfNYkf |
+|  ds001 | single task, multiple runs, in-plane T2, events, T1w, BOLD | mri |  |  |  | https://bit.ly/2H0Z6Qt |
+|  ds002 | multiple tasks, multiple runs, in-plane T2, events, T1w, BOLD | mri |  |  |  | https://bit.ly/2H0Z6Qt |
+|  ds003 | single task, single run, in-plane T2, events, T1w, BOLD | mri |  |  |  | https://bit.ly/2H0Z6Qt |
+|  ds005 | single task, multiple runs, in-plane T2, events, T1w, BOLD | mri |  |  |  | https://bit.ly/2H0Z6Qt |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,15 +1,65 @@
...
-A set of BIDS compatible datasets with **empty raw data files** that can be used for writing lightweight software tests.
+# bids-examples
+
+This repository contains a set of BIDS compatible datasets with **empty raw
+data files**. These datasets can be useful to:
+
+1. write lightweight software tests
+1. serve as an example on how a BIDS dataset can be structured
 
 ALL RAW DATA FILES IN THIS REPOSITORY ARE EMPTY!
```
effigies: It might be worth noting that, for the synthetic dataset, the NIfTI files are truncated at their headers so they can be loaded and their metadata read.

README.md
```diff
@@ -1,15 +1,65 @@
 [![Travis](https://api.travis-ci.org/bids-standard/bids-examples.svg?branch=master "Travis")](https://travis-ci.org/bids-standard/bids-examples)
 
-# BIDS-examples
-A set of BIDS compatible datasets with **empty raw data files** that can be used for writing lightweight software tests.
+# bids-examples
+
+This repository contains a set of BIDS compatible datasets with **empty raw
```
effigies:
```suggestion
This repository contains a set of BIDS-compatible datasets with **empty raw
```

README.md
```diff
@@ -1,15 +1,65 @@
...
-- You can find full versions of most MRI datasets in this repository [here](https://drive.google.com/drive/u/0/folders/0B2JWN60ZLkgkMGlUY3B4MXZIZW8).
-- You can find the full version of mne_sample_data [here](https://drive.google.com/drive/folders/0B_sb8NJ9KsLUQ3BMS0dxZW5nSHM).
-- You can find the full version of ds000246 (Brainstorm auditory tutorial) [here](https://openneuro.org/datasets/ds000246/versions/00001).
-- You can find the full version of ieeg_epilepsy (Brainstorm epileptogenicity tutorial) [here](https://openneuro.org/datasets/ds001779).
+To learn more about the Brain Imaging Data Structure (BIDS), visit
+http://bids.neuroimaging.io/
+
+If you want to contribute a dataset to the examples, we would be happy to
+accommodate it, as long as it exhibits some structure or particularity that is
+not covered by the already present examples so far.
```
effigies: This feels just a little weird to read. The meaning is clear, though, so feel free to reject.

```suggestion
not covered by the examples that are already present so far.
```

choldgraf:
The formatting looks good to me - I agree that the signal-to-noise is hurt by having so many datasets with overlapping structure. I'd rather something like:

1. Figure out the different patterns that should be highlighted, make sure there's one dataset that does this well.
2. Add links to other examples of the same dataset, or maintain a separate page that simply lists all of the datasets that are accessible (maybe just link to OpenNeuro?).

The challenge with the current setup is that the repo serves two purposes: one is to show what the BIDS structure is like, another is to list lots of different examples. I think these are two separate goals that deserve two separate pages. (though doesn't need to be in this PR)

README.md
```diff
@@ -1,15 +1,65 @@
...
+
+# Dataset index
 
-To learn more about the Brain Imaging Data Structure (BIDS), visit http://bids.neuroimaging.io/
+If you want to update this table, please open a new issue. For simple editing,
+this table is maintained in a [Google sheet](https://docs.google.com/spreadsheets/d/1d9KfmaKvTOpFWgi0Kd-tZGgVTsZk_4BHH9prcgFeigw/edit?usp=sharing), which is then converted to Markdown for this README file.
```
robertoostenveld: I would not keep the google sheet version available. For this initial formatting it is fine and helps a lot, but the google sheet and the markdown table on github are destined to drift apart, making automatic conversion from one to the other impossible.

README.md
```diff
@@ -1,15 +1,65 @@
...
 
-If you are looking for datasets with particular features, you can check out the documentation in the [WIKI](https://github.com/bids-standard/bids-examples/wiki/), which is sorted by imaging modalities.
+|  name | description | mri | meg | eeg | ieeg | full data |
+| --- | --- | --- | --- | --- | --- | --- |
+|  7t_trt | field maps, physiological data, quantitative T1 maps, T1w, BOLD | x |  |  |  | https://bit.ly/2H0Z6Qt |
+|  ds000117 | A multi-subject, multi-modal human neuroimaging dataset of 19 subjects on a MEG visual task | x | x | x |  | https://openneuro.org/datasets/ds000117/ |
+|  ds000246 | Auditory dataset used for Brainstorm’s general online tutorial | x | x |  |  | https://openneuro.org/datasets/ds000246/versions/00001 |
+|  ds000247 | Five minutes, eyes-open, resting-state MEG data from 5 subjects. This is a sample from The Open MEG Archive (OMEGA). | x | x |  |  |  |
+|  ds000248 | MNE sample data: Data with visual and auditory stimuli | x | x |  |  | https://bit.ly/2JfNYkf |
+|  ds001 | single task, multiple runs, in-plane T2, events, T1w, BOLD | x |  |  |  | https://bit.ly/2H0Z6Qt |
```
robertoostenveld: I wonder whether the datasets with the 2H0Z6Qt bitly link should also refer to openneuro. I think  that all of these originated from openfmri, but I don't know whether they have been curated/maintained there.

perhaps specify it as  
````
https://bit.ly/2H0Z6Qt or https://openneuro.org/datasets/dsXXXXX
````
for each of them (where XXX is e.g. ds001, although I now see that they have all been zero-padded, so we should link to the zero-padded URL)

franklin-feingold:
thank you @sappelhoff for raising this! I can help fill in the bit.ly for openneuro sets. We tried to preserve the ds assession number consistency when merging from openfmri to openneuro. I agree with @choldgraf for two different lists. The more categorized by type (with a coupe of representative examples) can be the more top level/easy to find and if a user needs more examples within a category then this can be linked out to the master list with the collection of all the datasets. The category classification may be challenging to determine the optimal level of abstraction. I think best in a different PR.

README.md
sappelhoff: good point :+1: the header is also intact for some EEG examples. I'll add this.

README.md
sappelhoff: yes, that's much better - thanks!

README.md
sappelhoff: good point :+1:

sappelhoff:
Thanks for the round of reviews everybody!

Let's move on and get this PR ready for merging and then tackle the next issue of "making the overview more concise" in a new PR.

@franklin-feingold I'll give you edit access to the [Google Sheet], and it'd be great if you could help me filling it up with the missing information / links to openneuro (instead of the bit.ly ones)

One additional thing: For the modality columns in the table, I'll change the `x` markings of when a modality is present to a repetition of that modalities name (`mri`, `eeg`, ...) --> this will make it easier to scroll down a long table and still know for each row, which modalities are present, without having to go back up to the header.

PR #130, opened 2018-10-22
ENH: Add models for ds003, ds114

effigies:
Not yet conformant with latest changes in [BEP 002](https://docs.google.com/document/d/1bq5eNDHTb6Nkx3WUiOBgKvLNnaa5OMcGtD0AZ9yms2M/edit?ts=5bcb2e82#).

ds000030 an ds000117, for which models either exist or are in-progress, are not represented in this repository.

Closes #129.

PR #160, opened 2019-05-18
ENH: Add models for ds003, ds114

effigies:
Replaces #130.

### DIFF
--- a/ds003/models/model-001_smdl.json
+++ b/ds003/models/model-001_smdl.json
@@ -1,0 +1,77 @@
+{
+  "name": "ds003_model001",
+  "description": "",
+  "input": {
+    "task": "rhymejudgment"
+  },
+  "blocks": [
+    {
+      "level": "run",
+      "transformations": [
+        {
+          "name": "factor",
+          "input": [
+            "trial_type"
+          ]
+        }
+      ],
+      "model": {
+        "HRF_variables": [
+          "trial_type.word",
+          "trial_type.pseudoword"
+        ],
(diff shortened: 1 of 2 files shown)


sappelhoff:
Thanks, you seem to have made this PR from your old fork ... are you sure that this will not reintroduce some bad history?

Why not delete your old fork, make a fresh one, and proceed?

effigies:
I rebased on the current master.

sappelhoff:
I see ... I also just realized that we won't introduce bad history via PRs, but rather via direct pushes to `master`.

Is this ready to be merged? If you need a review, you'd have to tag a member that has some experience with the models spec (I am not familiar enough with that).

effigies:
No, it's possible for someone to introduce bad history, but it should show up as a ton of commits in the PR. Typically just asking them to rebase on `master` will do the job. But re-forking is fine, too.

I'll request review from @tyarkoni and @adelavega.

tyarkoni:
I looked through the models (admittedly not very carefully, but short of trying to run them through pybids and fitlins, I don't know that my careful review is terribly helpful) and they LGTM.

sappelhoff (changes requested):
to get this PR merged, a `.SKIP_VALIDATION` file (like [in the derivatives ds](https://github.com/bids-standard/bids-examples/blob/master/ds000001-fmriprep/.SKIP_VALIDATION)) should be added to the ds003 and ds114 datasets, ... or the added `models` directory should be `.bidsignored`

PR #161, opened 2019-07-02
Adding example data for discussing EEG derivatives: face-13

Andesha:
Hi everyone,

As per the BEP021 post on the discussion group, I'm posting the eeg_face13 data set.

Our hope is to use this to explore what is necessary for EEG derivatives going forward in terms of ICA, annotations, and more.

Thanks!


eeg_face13/derivatives/lossless/sub-s01/eeg/sub-s01_task-faceO_eeg_qcr_annotations.json
```diff
@@ -0,0 +1,23 @@
...
+		"chan_bridge": "Pipeline decision flag indicating that channels were outliers for in terms of having high and invariant correlation coefficients to spatially neighbouring channels.",
+		"chan_rank": "Pipeline decision identifying the channel that has the least amount of unique information (highest average correlation coefficient to spatially neighbouring channels) to be ignored by ICA in order to account for the rank deficiency of the average referenced data.",
+		"comp_manual": "The interactive label (modified by analysts interactively or by a pipeline decision along with other labels) typically used to indicate which components are considered artifactual for any reason.",
+		"comp_ic_rt": "Pipeline decision flag indicating that components had a poor dipole fit (typically > %15 residual variance).",
+		"comp_brain": "ICLabel 0.3 Mark indicating that the compnents have cortical characteristics.",
+		"comp_muscle": "ICLabel 0.3 Mark indicating that the compnents have EMG characteristics.",
+		"comp_eye": "ICLabel 0.3 Mark indicating that the compnents have EOG characteristics.",
+		"comp_heart": "ICLabel 0.3 Mark indicating that the compnents have ECG characteristics.",
```
jasmainak: why do you prefer to use your own labels instead of the ones recommended in the draft specification? what was lacking?

eeg_face13/derivatives/lossless/sub-s02/eeg/sub-s02_task-faceO_eeg_qcr_annotations.json
```diff
@@ -0,0 +1,23 @@
+{
+	"Description": "Lossless state channel and component annotations.",
+	"IntendedFor": "derivatives/lossless/sub-s02/eeg/sub-s02_task-faceO_eeg_qcr.edf",
```
jasmainak: hmm ... I don't think the source suffix should be the part of the filename. Can you check?

jasmainak:
Looks like a good starting point! Thanks for making the pull request. I'd be curious to hear opinions from @sappelhoff @robertoostenveld @CPernet 

By the way, we can't include binaries / images in the repository. You can use empty file placeholders instead of them. The memory footprint should be as small as possible.

jasmainak:
Also, you are not passing the validator at the moment even for the source files. Might be nice to make it work so as to have a more focused discussion on the derivatives aspect.

sappelhoff:
Hi @Andesha, thanks for the contribution. I agree with Mainak: The dataset should be cleaned up a bit for now. Let me list some hints to help you:

# Working with continuous integration

To get more insights into the tests that are being run on Travis (our continuous integration service), click on the "Details link" 

image

Then search (using the browser function ctrl+f) for your dataset `eeg_face13`

image


And you can inspect the errors

# Big files

at the end of the Travis build, you can see whether big files were detected and should be deleted.

image

christinerogers:
Great to see this as a starting point for discussion! 
Would suggest: adding label "Work in progress" and editing PR title to include "face-13"

Andesha:
Hi everyone,

Thanks for the tips! It looks like I got everything working on the base BIDS side now.

In response to the question posed by @jasmainak above: the information listed in the label description contains our data processing specific annotations. This includes both channel and component marks. We use these measures to make decisions about whether to reject sources. The ICLabel marks come from the EEGLAB plugin of the same name found [here](https://sccn.ucsd.edu/wiki/ICLabel). This raises a good question of if we should signal if marks/annotations are user defined for their own processing or not.

Additionally @jasmainak is right. I believe the file suffix isn't really supposed to be there. However, we currently use this to differentiate between multiple sets annotations for the same derived file. In our case, "qcr" is post expert review.

jasmainak:
Excellent! One quick comment:

> we currently use this to differentiate between multiple sets annotations for the same derived file.

wouldn't you use the `desc` label to distinguish between them?

Andesha:
Would the file names then look like this?

`sub-s02_task-faceO_desc-qcr_annotations.tsv`

jasmainak:
yes exactly

Andesha:
Hi again,

I've just now updated the derivatives to take advantage of the desc label naming convention. Additionally I got rid of the source suffix in the sidecar files. Lastly, I moved three binary annotations from our "timeinfo" file and put them into the standard annotations tsv.

As it hasn't been discussed yet, we are currently using the "timeinfo" files to store our continuous annotations. This scheme is inspired by the BIDS definition of how to store continuous physiological data that is not part of a main data file, found [here](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/06-physiological-and-other-continous-recordings.html). A better name for this structure should be chosen in the future, but this is how the structure is named in our preprocessing techniques. Perhaps something signaling it is the continuous annotations belonging to a particular desc label.

Our main reason for this strategy is that this easily supports having multiple sets of annotations per derivative. Additionally, there would be no need to redefine a channel.tsv and create a new edf file for each annotation set.

Andesha:
Hi there,

Recently we've made significant changes to Face13. This includes adding more information at the raw bids level, as well as updating the annotation structure.

As such, I'll be closing this pull request and opening a new one shortly so we can have a fresh discussion.

Thanks!

PR #169, opened 2019-10-08
Fix TSV files UTF-8 encoding

gllmflndn:
Some TSV files do not appear to be UTF-8 encoded:
https://bids-specification.readthedocs.io/en/stable/02-common-principles.html#tabular-files
I applied this bash script:
```
for file in $(find . -name "*.tsv" -exec isutf8 -l {} \;);
do
  iconv -f ISO-8859-1 -t UTF-8//TRANSLIT  $file -o $file ;
done
```
I am not clear what should be used for microvolt, between µV, uV and microV.


robertoostenveld:
@dorahermes do you remember from the top of your hear the resolution for the "micro" in the uV units? It was allowed to use µV, right?

sappelhoff:
> It was allowed to use µV, right?

yes, see the [[appendix](https://bids-specification.readthedocs.io/en/stable/99-appendices/05-units.html)

actually, `uV` is NOT allowed to mean "micro volt". You can indicate "micro volt" by one of the following:

- `microV`
- `µV`

gllmflndn:
If so, then there are a number of datasets that need to be modified as they use `uV`:
- eeg_ds000117
- ieeg_epilepsy
- ieeg_visual_multimodal
- ieeg_visual

sappelhoff:
jep, these are errors that the validator does not catch currently. (unfortunately)

This would be a very nice and clear feature for the validator, see: https://github.com/bids-standard/bids-validator/issues/777

cc @rwblair

dorahermes:
sorry - this slipped trough around SfN time. At the time of the spec, we decided that it had to be µV. When writing these examples in Matlab for the iEEG datasets, I used char(181) for micro. Is this still correct? This keeps coming up and I would like to add a section in the starter kit.

gllmflndn:
@dorahermes: I think you should use `native2unicode(181,'latin1')` instead.

dorahermes:
Thank you! Just added a note in the bids-starter-kit (since this question keeps coming back), see
https://github.com/bids-standard/bids-starter-kit/wiki/FAQ#how-to-specify-units-in-microVolt

PR #171, opened 2019-10-18
Example for EEG derivatives: Face13

Andesha:
Hi all,

This pull request includes an updated version of the Face13 dataset. In terms of raw BIDS, several of the files have had more information added to them as to make the dataset more complete. On the derivatives side of things, the BIDS-Lossless-EEG data contains the current working state that we would like to use as a reference for the [Common Electrophysiological Derivatives](https://docs.google.com/document/d/1PmcVs7vg7Th-cGC-UrX8rAhKUHIzOI-uIOh69_mvdlw/edit#heading=h.gq2h70xj8umu) discussion. Specifically, ICA information, storage of data property annotations (both binary and continuous), multiple annotations for the same datafile (e.g. *desc-ll vs *desc-qc), and dipole information. 

Additionally, this version of Face13 contains our structure for working within nested BIDS derivatives (segmentation derivative within a preprocessed derivative) as outlined [here](https://github.com/bids-standard/bids-specification/issues/345#issuecomment-541211442).

Thanks!

PR #172, opened 2019-10-23
BEP021 Face13

Andesha:
Hi all,

As per [here](https://github.com/bids-standard/bids-examples/pull/171#issuecomment-544118651) and [here](https://github.com/bids-standard/bids-examples/pull/171#issuecomment-545502842), I'm reuploading the Face13 data set with derivatives to help drive discussion.

The issues previously outlined in the last PR still exist. For example, the derivatives need their own `dataset_description.json` files. I'll be giving BEP003 a close read shortly and fixing things ASAP.

Thanks!


Andesha:
I've updated the branch with the things mentioned in the previous pull request. I still need to give BEP003 and BEP021 another read through though!

robertoostenveld:
Just checking: Am I right that this PR combines the Face13 raw EEG dataset (in BIDS version 1.2.1 format), and that it also includes the derivatives (in an as of yet formally unspecified format)?

Did you pass the Face13 dataset without derivatives through the online validator?

robertoostenveld:
> Did you pass the Face13 dataset without derivatives through the online validator?

I suspect that was a silly question from me, Travis already did that for us, right?

Andesha:
Correct! I believe the raw is good to go

christinerogers:
Great to see this, @Andesha ! 

cc @CPernet  pas mal, check ca...

PR #174, opened 2020-01-01
ENH: Update ds00017 from version 1.0.0 to 1.0.3

apjanke:
I'm interested in updating the ds00017 example in bids-examples from its current version 1.0.0 to the new version 1.0.3 of the dataset. This new version includes derivatives, which will be useful for testing BIDS library code that supports derivatives (like BIDS-MATLAB may do in the future). This looks like the latest available version of ds00017.

I grabbed the version 1.0.3 dataset from https://openneuro.org/datasets/ds000117/versions/1.0.3/download and truncated all the data files using a custom script. I also removed what looked like a couple temporary files dropped by editors and included in the dataset by mistake, and removed the `.datalad` directory which contained application-specific config files and a `.gitattributes` that looks like it got pulled into the dataset.

This new version of the dataset is 3.6 MB after stripping. That seems right, since the old version is 2.7 MB.


apjanke:
The Travis run is failing due to multiple bids-validator exceptions on the new version of the dataset.

```
1: [ERR] Invalid JSON file. The file is not formatted according the schema. (code: 55 - JSON_SCHEMA_VALIDATION_ERROR)
		./dataset_description.json
			Evidence: .Funding should be array

	Please visit https://neurostars.org/search?q=JSON_SCHEMA_VALIDATION_ERROR for existing conversations about this issue.

	2: [ERR] Files with such naming scheme are not part of BIDS specification. This error is most commonly caused by typos in file names that make them not BIDS compatible. Please consult the specification and make sure your files are named correctly. If this is not a file naming issue (for example when including files not yet covered by the BIDS specification) you should include a ".bidsignore" file in your dataset (see https://github.com/bids-standard/bids-validator#bidsignore for details). Please note that derived (processed) data should be placed in /derivatives folder and source data (such as DICOMS or behavioural logs in proprietary formats) should be placed in the /sourcedata folder. (code: 1 - NOT_INCLUDED)
		./annex-uuid
			Evidence: annex-uuid

	Please visit https://neurostars.org/search?q=NOT_INCLUDED for existing conversations about this issue.

	3: [ERR] A json sidecar file was found without a corresponding data file (code: 90 - SIDECAR_WITHOUT_DATAFILE)
		./sub-emptyroom/ses-20090409/sub-emptyroom_ses-20090409_task-facerecognition_meg.json
		./sub-emptyroom/ses-20090506/sub-emptyroom_ses-20090506_task-facerecognition_meg.json
		./sub-emptyroom/ses-20090511/sub-emptyroom_ses-20090511_task-facerecognition_meg.json
		./sub-emptyroom/ses-20090515/sub-emptyroom_ses-20090515_task-facerecognition_meg.json
		./sub-emptyroom/ses-20090518/sub-emptyroom_ses-20090518_task-facerecognition_meg.json
		./sub-emptyroom/ses-20090601/sub-emptyroom_ses-20090601_task-facerecognition_meg.json
		./sub-emptyroom/ses-20091126/sub-emptyroom_ses-20091126_task-facerecognition_meg.json
	Please visit https://neurostars.org/search?q=SIDECAR_WITHOUT_DATAFILE for existing conversations about this issue.
```

I'll contact the dataset authors about this.

adelavega:
I'm guessing most of the failures were not caught in the original submission as the bids-validator has seem lots of recent updates/increase in coverage.
1) Can be fixed by making the "Funding" field an array containing a single string, rather than a string 
2) You should be able to remove `.annex-uuid` safely. This is also a datalad, git-annex specific file, as was not present in the original submission.
3) Looks like this error is occuring in the derivatives. This may be a validator error, as it looks like there *is* a corresponding file:

image

Maybe it's not checking for `.fif` files? Or these could perhaps not be BIDS-compliant?

rwblair:
@adelavega  @apjanke  For the meg.json files error 90, according to the spec it looks like the channels.tsv and meg.json need to be moved into the meg directory to be along side the .fif file instead of a directory above it:

https://bids-specification.readthedocs.io/en/stable/99-appendices/06-meg-file-formats.html#neuromagelektamegin

robertoostenveld:
It would be an option as well to use `.bidsignore` to deal with `annex-uuid`. See the validator docs for that. But in the example it could also be deleted.

For the channels.tsv and meg.json the [inheritance principle](https://bids-specification.readthedocs.io/en/stable/02-common-principles.html#the-inheritance-principle) applies. That does not mean (I think) that they can be at an arbitrary level, but metadata that is shared over multiple runs/sessions/subjects can be represented at a higher level. I believe that the `matching_pennies` example uses this.

rwblair:
Ah was being shortsighted by just looking at the specifc example instead of the spec at large.

Issue might be that the task name on the sidecars for sub-emptyroom and the .fif files are different? jsons are for task-facerecognition and .fifs are for task-noise.

sappelhoff:
@apjanke are you still interested in finishing your PR? What are the current blocking points?

apjanke:
I am interested in finishing this! But I'm a bit out of my depth here: are the remaining failures something I need to correct in my example files, or things that need to be tweaked in the test suite?

sappelhoff:
The validator and the underlying travis test script of this repo have both changed since you last opened this PR.

would you be comfortable with rebasing your changes on current master of bids-examples?

rwblair:
@apjanke I went ahead and rebased this branch on to master. The only issue remaining is the task name on the sidecar files in sub-emptyroom do not match the task name for the meg files. The sub-emptyroom directories and files look a bit different than what I'm seeing as current on openneuro.

sappelhoff (changes requested):
A `1.0.4` version of the data has been released recently: https://openneuro.org/datasets/ds000117/versions/1.0.4 including several fixes.

It'd probably best to:

1. download that 1.0.4 data
2. Make sure it passes validation (OpenNeuro says that it does, but better to check)
2. zero out the big files
3. update this PR to push version 1.0.4 instead ... then we'll end up with a perfect mirror between the bids-examples version and the openneuro version once more.

sappelhoff:
@apjanke I took over this PR and continue it in https://github.com/bids-standard/bids-examples/pull/223

I added you to the commit message as a co-author to credit your contribution. thanks a lot that you gave it a start many months back!

PR #176, opened 2020-02-05
ieeg_epilepsy_ecog: Fixes for new BIDS-validator rules (units symbols, Funding)

ftadel:


sappelhoff:
actually, the `microV` should be a valid combination and not necessary to change to `µV`. Did you get a validation error?

ftadel:
Yes, from OpenNeuro, on this dataset:
https://openneuro.org/datasets/ds001868

sappelhoff:
cc @rwblair --> I think microV should be a valid unit. It's combining a valid prefix ("micro") with a valid unit ("V"). see https://bids-specification.readthedocs.io/en/stable/99-appendices/05-units.html

fyi I am going to prepare a PR soon that is aimed at clarifying units (unicode, ascii, ...)

ieeg_epilepsy_ecog/dataset_description.json
```diff
@@ -12,6 +12,6 @@
     ],
  "Acknowledgements": "n/a",
  "HowToAcknowledge": "n/a",
- "Funding": ["n/a"],
+ "Funding": [],
  "DatasetDOI": "n/a"
-}
+}
```
sappelhoff, line 17:
```suggestion
}

```

PR #178, opened 2020-03-05
Bep 018 example genetics_ukbb

rwblair:
Dataset added by @CPernet as an example for bep 018. After discussing with @franklin-feingold we will merge this first and then validator so that validator may use example in its circle tests.


CPernet:
great, thank you -- let me know if anything else need change

PR #182, opened 2020-04-03
enable individual bids-validator config files for unfixable examples

rwblair:
Validator had required EchoTime1 and EchoTime2 to be defined for a phasediff.


genetics_ukbb/phasediff.json
```diff
@@ -0,0 +1,4 @@
+{
+   "EchoTime1": 0.0,
+   "EchoTime2": 0.001
```
sappelhoff, line 3: did you make these values up @rwblair ?

what's keeping us from finalizing this PR?

rwblair: yup these are made up. Checked with  dataset owner and these were not recorded when the dataset was created. There has to be a non zero difference between them to allow validator to pass.

I made one other deviation from the original dataset data to make it bids compatible for the example, Updated genetics_ukbb/task-FaceShapeEmotion_bold.json genetics_ukbb/task-rest_bold.json repition time to be 7.35 instead of 0.735 since there were slice timings upwards of 5 seconds. @sappelhoff You ok with this, I figured examples should be closer to the ideal of the standard than the idiosyncrasies of the datasets (that conflict with the standard) the examples were derived from.

ieeg_epilepsy_ecog/sub-ecog01/ses-postimp/sub-ecog01_ses-postimp_scans.tsv
```diff
@@ -1,3 +1,3 @@
 filename	acq_time
 anat/sub-ecog01_ses-postimp_T1w.nii.gz	1900-01-01T00:00:00
-ieeg/sub-ecog01_ses-postimp_task-seizure_run-01_channels.eeg	1900-01-01T00:00:00
+ieeg/sub-ecog01_ses-postimp_task-seizure_run-01_ieeg.eeg	1900-01-01T00:00:00
```
sappelhoff, line 3: tagging @ftadel to make him aware of this fix for his dataset

genetics_ukbb/phasediff.json
sappelhoff: > yup these are made up. Checked with dataset owner and these were not recorded when the dataset was created. There has to be a non zero difference between them to allow validator to pass.

is there no way to re-compute the values from the data?

>  Updated genetics_ukbb/task-FaceShapeEmotion_bold.json genetics_ukbb/task-rest_bold.json repition time to be 7.35 instead of 0.735 since there were slice timings upwards of 5 seconds.

so 0.735 was probably a typo?

--- 

I think we should try to not include made-up values. Worst case, could we include an exception for datasets where we have missing data through the bids-validator config :thinking:

rwblair:
Ok reverted repetition time. This ukbb is a multiband acquisition, 8x factor. If we multiply the TR by 8 we get 5.88 which is slightly less than the highest slice timing time, 5.63. https://biobank.ctsu.ox.ac.uk/crystal/crystal/docs/brain_mri.pdf (section 2.5) mentions a single band image is also taken. @CPernet  is there a chance that the slice timings in the genetics_ukbb/task-rest_bold.json and genetics_ukbb/task-FaceShapeEmotion_bold.json are for the single band image?

Only one dataset, ieeg_visual_multimodal, in the examples makes use of `MultibandAccelerationFactor` 

It has an acceleration factor of 3 and we see three distinct sets of slice timings in its task json files:
https://github.com/bids-standard/bids-examples/blob/master/ieeg_visual_multimodal/sub-som682/ses-nyu3t01/func/sub-som682_ses-nyu3t01_task-spatialpattern_run-04_bold.json#L82

genetics_ukbb/.bids-validator-config.json
```diff
@@ -0,0 +1,3 @@
+{
+    "ignore": [99, 15, 66]
+}
```
sappelhoff, line 3: This looks good to me @rwblair! Can we also have a `.bidsignore` in `genetics_ukbb` to ignore this file?

rwblair: Validator ignores dot files by default, it causing issues other places?

sappelhoff: true, see: https://github.com/bids-standard/bids-validator/issues/951

rwblair:
@sappelhoff you ok with the language/table I added to the README?

sappelhoff:
jup!

ieeg_epilepsy_ecog/sub-ecog01/ses-postimp/sub-ecog01_ses-postimp_scans.tsv
ftadel: @sappelhoff Thanks, I reported the fix on OpenNeuro (https://openneuro.org/datasets/ds002904/versions/1.0.1) and on the Brainstorm website.

PR #183, opened 2020-05-11
corrections to remove double suffix in _coordsystem.jon

dorahermes:
These corrections remove the double suffix in the derivatives surfaces and in the _coordsystem.json reference to this surface. 

Aligns with pull request on bids-standard ENH: tune suggested naming for the surface to follow BEP011 #463


ieeg_motorMiller2007/sub-bp/ses-01/ieeg/sub-bp_ses-01_space-ACPC_coordsystem.json
```diff
@@ -1,8 +1,8 @@
 {
-    "IntendedFor": "/derivatives/surfaces/sub-bp/ses-01/anat/sub-bp_T1w_pial.surf.gii",
+    "IntendedFor": "/derivatives/surfaces/sub-bp/ses-01/anat/sub-bp_pial.surf.gii",
```
yarikoptic, line 2: Note, as cited in https://github.com/bids-standard/bids-specification/pull/463#issuecomment-626796934 `_desc-T1w` is current recommended by BEP003 way to annotate the source "suffix" (modality), which could come handy here

dorahermes: Thanks @yarikoptic that will be a good option! Should I implement the _desc-T1w here already?

yarikoptic: one one hand: If examples should follow stock BIDS, then I guess `_desc-` better waits for common derivatives to be merged. **BUT** because stock BIDS I believe pretty much allows anything to fall under `derivatives/` subfolders -- you could as well add it ;)

sappelhoff: >  If examples should follow stock BIDS, then I guess _desc- better waits for common derivatives to be merged

I agree

>  BUT because stock BIDS I believe pretty much allows anything to fall under derivatives/ subfolders -- you could as well add it 

I think so too :-)

dorahermes:
I just updated this pull request with the _desc-T1w tag as suggested by @yarikoptic

PR #188, opened 2020-06-26
PET example data set

mnoergaard:
PET example data set in BIDS format for BEP009


robertoostenveld:
this is not ready for the master, I have merged this in the bep009_pet branch instead.

PR #191, opened 2020-07-17
ENH: Improve eeg_matchingpennies: 0.1.6 -> 0.2.0

sappelhoff:
The eeg_matchingpennis dataset is very straight forward and could be a good starting point to add some HED tags for example and validation purposes.

crossref: https://github.com/bids-standard/bids-specification/pull/527 and https://github.com/bids-standard/bids-validator/pull/1015

I have never HED-tagged a dataset, so this is just an attempt and I'll need @VisLab and @happy5214 to advise me on this.

I came up with this tag for the events in `events.tsv`:

`Event/Category/Participant response, Action/Button release, Experiment context/Sitting`

The tag describes the basic event that a participant either raises their left, or their right hand while staring at a fixation stimulus on a screen.

- Does this tag seem fine to you?
- How could this tag be further enhanced?
- Do you have other suggestions?

---

Full experiment description: https://github.com/bids-standard/bids-examples/blob/23ab390fdfb4d38b3e3eeabdbd663e2ac09a16bc/eeg_matchingpennies/task-matchingpennies_eeg.json#L3

pasted here for convenience:

> The task is emulating a game of 'matching pennies', where for each trial a subject and a computer agent both make a binary choice (here in the form of showing a left or right hand) at the same time.

> Crucially, one of the two players wins whenever the binary choice is equal, while the other player wins whenever the binary choice is not equal.

> Practically, subjects were seated in front of a computer screen on a desk with two computer mice glued to the desk at approximately 1 meter distance. The experiment consisted of three stages, separated by short breaks.

> Each stage contained 100 single trials. A subject started each trial in a self-paced manner by pressing down the outermost buttons of the two computer mice using their left and right ringfinger respectively while resting the hands calmly on the table. Pressing down both keys started a three-second countdown that was shown on the computer screen as a continuous shrinking of a horizontal bar.

> Throughout the countdown, subjects were focusing on a fixation stimulus in the middle of the computer screen. At then end of the countdown, subjects were supposed to quickly raise one of the two hands “as if to show the palm to the computer”. The movement was supposed to happen precisely timed with the end of the countdown.

> To ensure this, there was a 200 ms window at the end of the countdown during which the movement was required to occur. If the movement occurred outside this time window or if both buttons were released, the trial was restarted. Otherwise, the computer’s choice (left or right) was shown on the screen in the form of a left or right hand stimulus.

> If the subject’s choice was the same as the computer’s, this was a win trial for the computer. Otherwise, it was a win trial for the subject.

> This task was modeled after the task in Matthias Schultze-Kraft et al. “Predicting Motor Intentions with Closed-Loop Brain-Computer Interfaces”. In: Springer Briefs in Electrical and Computer Engineering. Springer International Publishing, 2017, pp. 79–90. doi: 10.1007/978-3-319-64373-1_8.


sappelhoff:
I just found another rather obvious tag to add: `Paradigm/Matching Pennies game` ... so the whole one would be:

`Paradigm/Matching Pennies game, Experiment context/Sitting, Event/Category/Participant response, Action/Button release`

VisLab:
Stefan,
We have been avoiding the Paradigm tags. They are too ambiguous and turn
out not to be useful for machine learning.  I will send you a list of
questions in the next day or two. I need to download the example and really
try to understand the dataset. The first step is to have a good text
description of each event type and to assign the event category.
Kay

VisLab:
I have looked at the tagging and I think it would be useful to have a back and forth on this before the pull request is merged. The description is somewhat vague.  Sometimes it refers to buttons and sometimes to keys.  It seems that the original experiment used the keyboard, but this experiment used two mice?

I looked at the event files and couldn't understand what the numbers meant. What exactly did the onset event mark? was it the very start before the countdown. If so, why are the latencies negative?   What are their units. If in seconds, this isn't nearly enough resolution to do any meaningful analysis.

The experiment had as a goal measuring motor intent. That needs to be clearly defined in the tags. 

The other big issue here is how to define events so that the dataset can be analyzed with out an extreme amount of manual coding. I need to actually understand what the events in the .tsv file and what the other columns defined relative to the onset and what units they are in before I can make progress.  It might be that we want an expanded version with more events in the record rather than these columns with values that you need to compute relative to.

sappelhoff:
Thanks a lot for looking into the data.

>  It seems that the original experiment used the keyboard, but this experiment used two mice?

Yes, two mice glued to the table at a certain distance

> I looked at the event files and couldn't understand what the numbers meant.

it's in the `task_matchingpennies_events.json` file, see this excerpt:

```JSON
    "value": {
        "Description": "The code (EEG Marker value) associated with an event ",
        "Levels": {
            "1": "Release of left button (i.e., raising left hand)",
            "2": "Release of right button (i.e., raising right hand)"
        }
```

So the listed events correspond to the button release.

>  What are their units. If in seconds, this isn't nearly enough resolution to do any meaningful analysis.

The units are described in `task_matchingpennies_events.json` as well. `response_time` is in milliseconds, `onset` in seconds, and `latency` in milliseconds

> The experiment had as a goal measuring motor intent. That needs to be clearly defined in the tags.

yes, which tag would that be? Would it be some tag in the **Paradigm** group taken from http://www.hedtags.org/display_hed.html ?


> It might be that we want an expanded version with more events in the record rather than these columns with values that you need to compute relative to.

I see your point. Perhaps this dataset is not very well suited to serve as an example after all.

This data was collected as a part of a student project and it was converted to BIDS some time early in 2018. Now that we are looking into this data with the HED angle, I see that it's problematic that several events are somehow "collapsed" in the events.tsv file.

For example, the rows in events.tsv with their EEG trigger values of 1 or 2 refer to the "button release" in the experiment ... in the real experiment, there was a delay after the button release and then the stimulus was shown (a left or a right hand picture). Yet in the events.tsv file, it seems as if it all happened simultaneously.

I realize now that that's certainly not the best practice, and not a part where this example shines.

We can also decide to not HED tag this example and pick another one.

VisLab:
Let's explore the data a little further.  As it is, I don't think it is usable for analysis.  Do you know what the units are of the other columns? Can you draw a picture of the events represented in the experiment relative to the numbers in these columns?  

If you can, and the resolution is sufficient, we could write a program to remap the events into something that is usable and then document.  I'm willing to consider doing it, but I could also look at one of the datasets that I am planning to upload in BIDS format and see if one of them has a simple enough event structure to be a good example.  

The fivepennies example is instructive in that it illustrates a fundamental problem with data-sharing for EEG and related technologies such as MEG.  Most analysis for these datasets is event centric. If careful thought isn't given as to what the events are that are included, the dataset won't be useful to others.

sappelhoff:
>  Do you know what the units are of the other columns

all units are described in [this file](https://github.com/bids-standard/bids-examples/blob/master/eeg_matchingpennies/task-matchingpennies_events.json). Not all columns have units --> some are boolean values, others are factor variables with some levels. But everything is defined in the linked file.

> Can you draw a picture of the events represented in the experiment relative to the numbers in these columns?

The events in the experiment were like this:

- throughout the experiment there is a central fixation stimulus (combination of bulls eye and cross)
- participant places both ring fingers on the outer mice buttons each (left and right)
- this starts a countdown (visual on the screen: shrinking horizontal bar). duration: 3secs
- after the countdown is done, participants have to react by raising the left or right hand
- after raising a hand, there is a brief time with only the fixation cross (~60ms on average) as the BCI prepares the feedback
- then a drawing of a hand is presented on the left or right of the screen
- then there is only the fixation cross, and the participant can start a new trial by again placing both ring fingers on each mice button (left and right)

(not a drawing, but I hope it'll do :slightly_smiling_face: )

Finally:

I dug out an old harddrive and inspected the experiment code from when this data was recorded. This is what I could find:

- The fps of the presenting screen were 60Hz, which explains that the entries in the `response_time` column are multiples of 1/60 (16.66666) milliseconds
- The only EEG triggers that were being sent are in the `value` column (there aren't any more hidden, undocumented trigger values ... that's all we have). Namely a `1` upon left hand button release, or a `2` upon right hand button release.
- The countdown was 3 seconds ... *after* the end of the countdown, participants had to raise one of their hands ... upon registering this "hand-raise", a trigger (`1` or `2`) was sent, and the reaction time was calculated online, and saved. Reaction time here refers to the interval from end of the countdown until raise of one hand (=button release) ... this is encoded in the `response_time` column.
- After the button release (event with value `1` or `2`), there was some time X and then a stimulus was shown (drawing of a hand) either on the left or right of the screen for a duration of 1 second. the time `X` cannot be exactly reproduced, but it's something like this: `50ms + ~10ms + ~8ms`, where 50ms were a hardcoded break before getting the online data from LSL, ~10ms was the time to get a prediction from the BCI classifier, but this varied around 10ms, and finally the ~8ms at the end come from the uncertainty of the refresh rate of the screen (60fps = 1 screen every 16.66666ms)


> . Most analysis for these datasets is event centric. If careful thought isn't given as to what the events are that are included, the dataset won't be useful to others.

indeed, I absolutely agree.

sappelhoff:
One more comment from my side though: Even if this example cannot be a perfect display of the power of HED tags, it could still serve as a valuable test case.

- We can tag this dataset so that it passes HED validation (tags won't be perfect, or helpful for analysis ... but at least valid)
- We add the new `HEDVersion` to the `dataset_description.json`

... and then the HED validation should always pass, even if HED schema is updated in the future.

VisLab:
I agree this would be useful.  I'm going to try to understand the dataset a
little more before suggesting a specific strategy.

VisLab:
Okay..... I think I have a better picture.  Here are my comments pertaining to _events.tsv and _events.json.

1)  The onset column in _events.tsv has been rounded to the nearest second.  Since EEG analysts will assume that they should time-lock to the onset, this makes it impossible to do analysis.  Luckily there is a sample columns --- so we could rewrite the files to put real onsets.  For example the first onset should probably be 18.1554 not 18. 

2) I think there is enough information in the file to rewrite the file to expand the events into:
      1) Countdown starts
      2) Countdown ends
      3) Button released (the only event that is currently in the file)

3)  We need to decide about how to express the matching of the hands.  There are 4 conditions:
       participant raises right hand  (and it matches)
       participant raises left hand  (and it matches)
       participant raises right hand (and it doesn't match)
       participant raises left hand (and it doesn't match)

What is intrinsically important that people might want to analyze?   Well first is the motor action and choosing which hand to raise.  That is encoded by the trial_type.   Then later there will be a neural response when the user sees whether it matched or doesn't. That neural response will occur sometime after the X time which is the display of the image.  My thought is that the estimated time that you mentioned above should be documented, but we should put actual events in for it because of the uncertainty in timing.

4)   Could you point me to the place in the spec where the Levels are specified as a field for _events.json?  I know I saw it somewhere, but I can't find it in the current spec.  

My thought is to rewrite the events and use HED tag description instead of levels.  It doesn't seem that the interaction of Levels and HED "codes" is well-specified. I think they could be used in conjunction.

VisLab:
Could you bring some more people into this discussion?  I am concerned that without some refactoring this example might not lead to good practice for EEG data-sharing.  Also, this might be an opportunity to clarify how events should be documented in the BIDS spec.

sappelhoff:
>  so we could rewrite the files to put real onsets.

agreed ... BUT I think that EEG analysts should always prefer the `sample` column, because that's an integer index into the data and thus the most accurate way to get the true onset.

> We need to decide about how to express the matching of the hands.

This is already encoded by comparing the `trial_type` and `pred` columns in the `events.tsv` files.

To follow your convention:

participant raises right hand (and it matches) -->`trial_type=right` and `pred=right`
participant raises left hand (and it matches) --> `trial_type=left` and `pred=left`
participant raises right hand (and it doesn't match) --> `trial_type=right` and `pred=left`
participant raises left hand (and it doesn't match) --> `trial_type=left` and `pred=right`



>  My thought is that the estimated time that you mentioned above should be documented, but we should not put actual events in for it because of the uncertainty in timing.

agreed

> Could you point me to the place in the spec where the Levels are specified as a field for _events.json?

I am not entirely sure what you mean by "Levels", but perhaps you meant this part of the spec (and especially the example):

https://bids-specification.readthedocs.io/en/latest/02-common-principles.html#tabular-files

> My thought is to rewrite the events and use HED tag description instead of levels. It doesn't seem that the interaction of Levels and HED "codes" is well-specified. I think they could be used in conjunction.

I don't understand what you mean here, which is going back to my point before that "I am not entirely sure what you mean by Levels"

> Could you bring some more people into this discussion?

yes, we could ask @robertoostenveld who may be interested once he is back from vacation

but we can bring in all sorts of people when we have a concrete question that we want advice (or a 2nd opinion) on --> but currently I don't see a disagreement either between ourselves or with the spec. (feel free to tag anyone you'd like to chime in though, this is all public anyways)

---

Concrete take aways

- [ ] update README with a brief note that onset of feedback (hand stimulus) was not recorded, but can be estimated to be around XYZ
- [ ] re-calculate onset times based on `sample` column and do not round the values
- [ ] potentially add two new kinds of events: Countdown start, and Countdown stop --> Yet we do not have the exact `sample` of when this occurred and there is probably an uncertainty of  ~8ms due to the refreshrate of the screen (60Hz)

VisLab:
Yes, we should definitely recalculate the onset. Since onset and duration are the only required fields, most code will use that as the time of the event.

I think that the Bids spec should be updated in two places:

1)  The section on 05-task-events has an example where it shows a sample events file with a trial_type column.  It should include a json sidecar below it with the Levels defined for the trial_type. Researchers should be encouraged to provide sidecar information

2)  The appendix on hed should be updated to show a json sidecar where my_codes has both Levels and HED descriptors.  They should not be mutually exclusive. The current specification doesn't satisfy the common principles.  The my_codes should have a LongName, Description, etc.   since it corresponds to a name of a column.

VisLab:
One other question about pennies.....

Why isn't the response_time column given as negative values?  It is relative to the onset right?

sappelhoff:
True, but if you look at the definition of that column, it says:

```JSON
    "response_time": {
        "Description": "Reaction time to lift the left or right hand after the countdown reached zero.",
        "Units": "Milliseconds"
    },
```

But I agree with you that a negative response_time would perhaps be more easily understandable.

VisLab:
It seems like that the response_time was expressed relative to the onset?  Even after our discussion, I still don't understand it since it seems that the ONLY events that we actually have times for are the times that the button was released (the onset)?   Could you go over one more time what those numbers stand for and how they were obtained?

On the match issue ---- to capture the intent of the experiment, we have to encode that is it is not just which hand was raised and which was displayed, but whether the hands matched.  This could be encoded in the events.tsv HED column, as could some of the latencies if we choose.   

As we discussed elsewhere, the common_principles says that all tsv columns should be identified in a json sidecar.  I think once we settle on everything, the sidecars should also be created.... possibly just one at the top level.  After all this is supposed to be a model for people to use.

sappelhoff:
> Could you go over one more time what those numbers stand for and how they were obtained?

Yes, the deciding word is **online** see my previous explanation with added emphasis:

> The countdown was 3 seconds ... after the end of the countdown, participants had to raise one of their hands ... upon registering this "hand-raise", a trigger (1 or 2) was sent, and the reaction time was calculated **online**, and saved. Reaction time here refers to the interval from end of the countdown until raise of one hand (=button release) ... this is encoded in the response_time column.

Think in terms of experimental code:
- the shrinking of the error bar is done - on the screen, now nothing happens ... participant is supposed to raise a hand
- internally, psychopy software is counting the number of screen refreshed (one ever 16ms) and checking at each refresh, whether the participant has raised a hand
- as soon as a hand raise is registered, the counted number of screen refreshes is converted to seconds ... and saved

--> that's the `response_time` column

> As we discussed elsewhere, the common_principles says that all tsv columns should be identified in a json sidecar. I think once we settle on everything, the sidecars should also be created.... possibly just one at the top level. After all this is supposed to be a model for people to use.

this file already exists, please see again: https://github.com/bids-standard/bids-examples/blob/master/eeg_matchingpennies/task-matchingpennies_events.json

of course when we change events.tsv, the existing events.json will have to be adjusted / extended.

VisLab:
Thanks......... so the response time is:   Onset - (End of countdown)?

The goal of HED is to provide information for the events that makes the
information understandable for analysis in a somewhat generic way so that
analysts don't have to rewrite their code and bother the researchers with
all these questions.

HED has two types of fields --- the instance-specific information should
appear in the HED column in the .tsv.   The event-type specific (as
encapsulated by the trial_type categories) should appear in the json
sidecar. Analysis tools should union all these tags. This example could
illustrate both types as follows:

Since all of the events are essentially of one or two similar types: a
participant response that is a mouse release and raising of one hand or the
other, those tags should be in the JSON sidecar.

In the HED column  in the tsv I would put two tags in each row:   One
encoding the response time with the units and the other encoding whether it
was a match or not.

We could also add tags in JSON side car to encode what stage meant.  This
is an experimental condition.  HED-3 (out in September) makes it easier to
encode those, but it is still doable now.

We would have to write a script to convert the existing .tsv files to
extract the response and match information from the other columns.  We
would need to write a script anyway to convert the onsets to an unrounded
form anyway.

What do you think?

VisLab:
You know, the tools that expand the annotations for the events could be
written  so that the columns like the response_time could also have their
HED annotation just be put in the JSON sidecar as:

/Attribute/ResponseStartDelay/# ms

with a specification that the when a # appears that the corresponding
column value in the .tsv file should be expanded.  We are in the process of
doing some library functions that do the expansion anyway, we could just
include that if it is in the bids specification. The HED spreadsheet
validator already allows users to specify which columns have HED tags and
to allow certain columns (such as a column containing the description) to
have HED tags expanded so they don't have to appear in the spreadsheet.  I
don't think this will affect the bids-validator except we have to make sure
the # is correctly handled.

sappelhoff:
> What do you think?

I'll sit down soon and write some code that will enrich the events in eeg_matchingpennies.

And then we can add HED tags!

Specifying the tags both in JSON and TSV and then analyzing the union still feels a little odd to me ... but I trust your judgment that this is a convenient workflow.

> /Attribute/ResponseStartDelay/# ms

this sounds like a convenient syntax - but I would consider proposing it separately from the issues we are currently dealing with.

sappelhoff:
In this commit, you can see the reformatting script that I wrote: https://github.com/bids-standard/bids-examples/pull/191/commits/765e678e1b0e8dcc5589f5043568a5ed2b091e6d

note that I will later remove it from the PR --> it's not going to be merged anywhere, I just post it here for our convenience and so that you can see approximately what I did

In this commit, the changes that resulted from running the script are documented via git diff: https://github.com/bids-standard/bids-examples/pull/191/commits/8f9b8830a80836048506848edabdcbe0f499c671

What do you think? Is something missing? could something be improved? --> else, next step would be adding the HED tags.

eeg_matchingpennies/sub-05/eeg/sub-05_task-matchingpennies_events.tsv
```diff
@@ -1,301 +1,901 @@
...
-1819	0	left	83.3333333333	right_hand.png	1	9095824	3	93	left	1	19	-2.2
-1823	0	right	50.0	right_hand.png	2	9118763	3	94	left	1	19	-12.0
-1828	0	left	50.0	right_hand.png	1	9141696	3	95	left	1	19	-11.8
-1832	0	right	50.0	right_hand.png	2	9164627	3	96	left	1	19	-15.6
-1837	0	right	83.3333333333	left_hand.png	2	9187727	3	97	right	1	19	-27.6
-1842	0	left	100.0	right_hand.png	1	9210913	3	98	left	1	19	-5.6
-1846	0	right	116.666666667	left_hand.png	2	9234202	3	99	right	1	19	-29.8
-1851	0	left	116.666666667	right_hand.png	1	9257470	3	100	left	1	19	-24.2
+onset	duration	trial_type	response_time	stim_file	value	sample	stage	trial	bci_prediction	bci_prediction_valid	n_repeated	latency
+15.038933333332999	n/a	right	116.666666667	left_hand.png	n/a	n/a	1	1	right	1	2	-4.2
```
sappelhoff, line 2: as you can see, the onset column is now at full precision (not rounded anymore)

eeg_matchingpennies/sub-05/eeg/sub-05_task-matchingpennies_events.tsv
```diff
@@ -1,301 +1,901 @@
...
+18.038933333333	n/a	right	116.666666667	left_hand.png	n/a	n/a	1	1	right	1	2	-4.2
+18.1556	0	right	116.666666667	left_hand.png	2	90778	1	1	right	1	2	-4.2
+19.8922	n/a	right	100.0	right_hand.png	n/a	n/a	1	2	left	1	2	-2.4
+22.8922	n/a	right	100.0	right_hand.png	n/a	n/a	1	2	left	1	2	-2.4
+22.9922	0	right	100.0	right_hand.png	2	114961	1	2	left	1	2	-2.4
+24.5304666666667	n/a	left	33.3333333333	right_hand.png	n/a	n/a	1	3	left	1	2	-17.2
+27.5304666666667	n/a	left	33.3333333333	right_hand.png	n/a	n/a	1	3	left	1	2	-17.2
+27.5638	0	left	33.3333333333	right_hand.png	1	137819	1	3	left	1	2	-17.2
```
sappelhoff, line 10: look at the three rows above --> they show a pattern that repeats throughout the events tsv files:
- countdown onset
- countdown offset
- button release

all belong to a single trial ... but have different `sample` values ... we only know the sample for "button release". etc.

eeg_matchingpennies/task-matchingpennies_events.json
```diff
@@ -39,28 +39,28 @@
...
+    "bci_prediction": {
+        "Description": "Which lift_side did the BCI predict in this trial.",
         "Levels": {
             "left": "Left",
             "right": "Right"
         }
     },
-    "pred_valid": {
-        "Description": "Did a prediction arrive on time (if not, a random prediction was made)",
+    "bci_prediction_valid": {
```
sappelhoff, line 56: I renamed some columns to make them more easily understandable.

sappelhoff:
I think I'd take this PR as a standalone now to upgrade eeg_matchingpennies from version 0.1.6 to version 0.2.0.

Then in a next PR we can format it from 0.2.0 to 0.3 by adding HED tags. (taggin fyi: @VisLab )

VisLab:
This may not be the right place to ask this, but do we have any examples of BIDS datasets with derivatives that also have events?  I would like to better understand how events will migrate through the derivatives and what information will be included.  I have been rereading the BIDS spec more carefully.  

The definition of events seems to only allow stimulus and response events.  Most EEG records a lot more events than that.  Things like instructions and pausing to take surveys may not be considered "stimuli" but may be important in continuous analysis or for long-term monitoring applications.  Also, many analyses want to insert events marking things that are detected by the algorithms --- such as a place where a channel drops out or where there is a blink or saccade.

sappelhoff:
> do we have any examples of BIDS datasets with derivatives that also have events?

I assume you are talking about the new "common derivatives" that are new with the BIDS 1.4 release --> to my knowledge we don't have examples for that yet (still to be done)

> The definition of events seems to only allow stimulus and response events.

To my knowledge any event can be included, what makes you think only stimuli and responses are permitted?

All the examples you mentioned can be encoded in events tsv as far as I see it

VisLab:
My confusion is from the description of event in common principles:

*Event* - a stimulus or subject response recorded during a task. Each event
has an onset time and duration. Note that not all tasks will have recorded
events (e.g., “resting state”).

sappelhoff:
I see, confusing indeed. I see no barrier to having events other than stimuli or responses --> feel free to open an issue on bids-specification about this and we'll see whether other people agree as well.

sappelhoff:
I am going to merge this now, but I'll have to update 

1. the data on OSF: https://osf.io/cj2dr/
1. the datalad mirror on github: https://github.com/sappelhoff/eeg_matchingpennies

Let's see when I get to that.

Thanks for the fruitful discussion so far @VisLab

Issue #192, opened 2020-07-20
electrodes.tsv mismatch with channels.tsv in iEEG examples


dipterix:
In example `ieeg_visual/sub-01/ses-01/ieeg/`

Channel names in `channels.tsv` are `iEEGxxx` while names in `electrodes.tsv` are `xxx`. Should two files be consistent? If not where can I find information to match these two files? 

For example, I select electrodes based on electrodes.tsv, but then I want to get information from channels.tsv. One easiest way is to have a one-to-one or one-to-many mapping so that I can join these two tables.

dorahermes:
Thank you for catching this! The two files should actually be consistent in naming. There was a 1-to-1 mapping in these data, so I corrected it accordingly with pull request [#193](https://github.com/bids-standard/bids-examples/pull/193)

dipterix:
I saw your PR, thanks for the information. I have another question unrelated to this issue:
In `sub-<label>_[ses-..]_task-xx[_run-..]_ieeg.json`, I didn't see any entry indicating where the data file is in BIDS specification document. Is it always save to assume changing extension guarantees the data files?

For example in `ieeg_visual`, there is `sub-01_ses-01_task-visual_run-01_ieeg.json`. Assuming we are using BrainVision format, if I change `.json` to `.vhdr`, can I guarantee the header file?

A follow-up issue is in `.vhdr` file, the meta information actually indicates data file name, i.e. `.eeg/.dat` files. The file name might not look like `sub-xxx_ses-xxx_task-xxx...._ieeg.eeg`. Should I stick to BIDS convention, or should/can I use data file indicated by header file, or both? And if both, how can I get that information?

sappelhoff:
> Is it always save to assume changing extension guarantees the data files?
> if I change .json to .vhdr, can I guarantee the header file?

not sure if it's guaranteed to work ... seems not very robust to do it this way. Have you looked at 

- github.com/bids-standard/bids-matlab
- github.com/bids-standard/pybids

? :-)

> A follow-up issue is in .vhdr file, the meta information actually indicates data file name, i.e. .eeg/.dat files.

that's correct ... the BrainVision file format consists out of three files ... and the VHDR and VMRK files both contain internal links to the other files (including the EEG file).

You will need to fix the internal links when renaming the data files to fit the BIDS convention.

You could do that manually ... but it's better to write a script for it. Or even better, to re-use software that's already out there to take care of this.

[...]

PR #193, opened 2020-07-20
corrected _electrodes.tsv files to match _channels.tsv

dorahermes:
surpasses #183 and thus closes #183 

closes [#192](https://github.com/bids-standard/bids-examples/issues/192)

also closes https://github.com/dorahermes/bids-examples/pull/1


dorahermes:
Thanks @sappelhoff ! I corrected the files in this pull request. I checked the channel names in the empty data files to make sure they match the _channels.tsv, which is why sub-01 and sub-02 differ.

Hopefully fixes issue #192  

I also made corrections to the _coordsystem.json and naming of the surface files in the derivatives. 

This passes the validator and has the extra lines at the end of the .tsv files.

dorahermes:
This pull request surpasses the previous one (#183)

I used Xcode or Atom to edit the json files, not sure why it removes the new line at the end. It is a bit tricky for me to check this, since these text editors do show a new line, for example, this is what the ieeg_visual/dataset_description.json file looks like to me in Atom:
image
Do you maybe know any way for me to check this?

sappelhoff:
Yes, you can go to "Packages" in atom and search for the "whitespace" package, which is installed on Atom by default.

Then go to "settings" for that package, and you'll see the following:

image

In this screenshot you also see my own settings, perhaps it'll show where our settings differ.

dorahermes:
Thanks @sappelhoff !
I have the same setting on Atom, I think it is correct in the current pull request.

sappelhoff:
Cool, thanks for all the fixes @dorahermes.

The newlines are still different, but that will have to remain a mystery. We don't want to keep opening and closing files just to make git happy :-)

dorahermes:
Thanks! (I did notice that there was still a change, but I think that the previous files actually did not have newlines at the end and the new ones do, the first set of json files were written using jsonio in Matlab and that workflow does not add a newline at the end)

sappelhoff:
you are absolutely right @dorahermes ... thanks for making me aware of it!

PR #196, opened 2020-08-04
ieeg_epilepsy_ecog: New OpenNeuro link 

ftadel:
The anatomical scans were not defaced in the original dataset. 
The dataset was deleted by the OpenNeuro team and reuploaded with a new dataset number.


sappelhoff:
thanks for the update @ftadel !

PR #197, opened 2020-08-11
Fix issues with eeg/ieeg/meg datasets.

### LINKED ISSUE #186: Cleaning up BIDS examples that showcase use of physiological and other continuous recordings
Dear @bids-maintenance and @robertoostenveld ,

we are finalizing the BIDS PET examples and we are using the physiological and other continuous recordings as inspiration (https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/06-physiological-and-other-continuous-recordings.html) to handle the blood data. 
Now we noticed that the examples that are uploaded that use the _physio.tsv are e.g. missing the associated _physio.json. I get that this is not required, but optional, but I can't seem to find an example where this file is present. 

Can we maybe add this to one of the examples?

Cheers,
Mel

rwblair:
closes #186


sappelhoff:
Thanks @rwblair --> I'd say this closes #186 and I edited the OP to make it automatically close that issue upon merge.

eeg_rishikesh/sub-001/ses-01/eeg/sub-001_ses-01_task-meditation_channels.tsv
```diff
@@ -76,5 +76,5 @@ GSR2	GSR	n/a
 Erg1	MISC	n/a
 Erg2	MISC	n/a
 Resp	RESP	n/a
-Plet	PPG	n/a
+Plet	PD	n/a
```
sappelhoff, line 79: are you sure this is a photodiode @rwblair ? also tagging @arnodelorme to make him aware that the meditation dataset needs an update because the PPG "type" is not a valid type.

What kind of channel was that Arno?

ieeg_visual/sub-01/ses-01/ieeg/sub-01_ses-01_task-visual_run-01_channels.tsv
```diff
@@ -1,119 +1,119 @@
 name	type	units	low_cutoff	high_cutoff	reference	group	status
-1	ECoG	µV	300	0.5	intracranial	1	good
```
sappelhoff, line 2: @dorahermes does this dataset need to be updated somewhere? i.e., the "real" dataset? --> just to keep the real data and the examples in sync

eeg_rishikesh/sub-001/ses-01/eeg/sub-001_ses-01_task-meditation_channels.tsv
robertoostenveld: A PPG is not simply a photodiode. For a PD (as per [IEEG specification](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/04-intracranial-electroencephalography.html#channels-description-_channelstsv)) I would assume that the channel represents the signal of a photodiode that is attached to the corner of a screen. That is common in clinical iEEG experiments in EMUs (Epilepsy Monitoring Units), where screens used for stimuli are often not precisely calibrated.

Would it not be better to add `PPG` as an allowed type? I would say that a PPG channel is unambiguous. We also use them often for physio recordings during fMRI sessions.

sappelhoff: I have never used `PPG` but I am happy to follow your and @arnodelorme's  lead here @robertoostenveld 

https://github.com/bids-standard/bids-specification/issues/561

A Pull Request is welcome, I'll be on vacation starting from tomorrow though.

ieeg_visual/sub-01/ses-01/ieeg/sub-01_ses-01_task-visual_run-01_channels.tsv
dorahermes: We've only shared the non-BIDS version currently, we plan to put the BIDS data on openneuro. Will make sure to  include these updates!

PR #201, opened 2020-09-21
New dataset: ds000001-fmriprep derivatives dataset

### LINKED ISSUE #187: Example derivatives sets, ideally for each modality and existing bids-app?
Relatedly, how can uploading a minimal set of example derivatives be systematized and included in the process of BIDS app submission? Or is it already?

@dPys

effigies:
closes #187 

cc @rwblair

### DIFF
--- a/ds000001-fmriprep/.bidsignore
+++ b/ds000001-fmriprep/.bidsignore
@@ -1,0 +1,10 @@
+*.html
+logs/
+figures/
+*_xfm.*
+*.surf.gii
+*_boldref.nii.gz
+*_bold.func.gii
+*_mixing.tsv
+*_AROMAnoiseICs.csv
+*_timeseries.tsv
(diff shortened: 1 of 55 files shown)


sappelhoff:
We have tons on mri datasets that all look pretty similar (correct me if I am wrong), could we make one of those existing dataset into a derivatives example instead of adding a new example? Renaming would be fine as well IMHO.

effigies:
I'm not sure I understand your suggestion. We have ds001, so this seemed like an appropriate addition.

robertoostenveld:
I would not have guessed that `ds001` is the same as `ds000001`. Can it be called  `ds001-fmriprep` for internal consistency in the examples?

effigies:
Sure.

sappelhoff:
I am commenting from mobile, sorry for being a pain.

I just think that instead of creating a new top level directory, it'd be nice to simply update the existing dataset. 

It may be that the situation is different from what it seems given the limited mobile ui. In that case, please ignore my suggestion.

sappelhoff (approved):
I obviously did not review each file, but I am now (after the discussion above) +1 to merge this.

It'll not pass validation for now, but it will help to extend the bids-validator to eventually validate derivatives.

PR #202, opened 2020-09-23
Update README.md

ftadel:
Dataset ds002904 had to be removed because it was not complying with OpenNeuro requirements. Now distributing it from the Brainstorm web server.


robertoostenveld:
Hi @ftadel I am curious what the compliance issue was. Did it have to do with permissions for reuse?

ftadel:
There was no hope that the hospitals (Grenoble and Freiburg) would accept distributing patient data under a CC0 license, or at least the neurologists who did the recordings didn't want to spend time fighting for it.
In Freiburg, I think they had a limited consent from the patient, not compatible with a public distribution. In Grenoble, they were not sure, so they decided to copy the Germans.

Since it doesn't impact in any way the Brainstorm tutorials we developed with both sites, they didn't have much interest in making efforts to keep the data on OpenNeuro. So these two nice intracranial datasets are gone...

robertoostenveld:
Yup, makes sense. I also consider CC0 to be suboptimal as a license for data from human participants and hope that OpenNeuro broadens the scope of data use agreements.

PR #207, opened 2020-11-04
ENH: Regularly run tests, run test for stable and master validator

sappelhoff:
With this PR I suggest that we:

- change from Travis CI to GitHub actions
- run the bids-examples CI suite once a week via a cron job to check even when there were no PRs or commits (every Monday, 4am)
- run the CI suite on both, the "latest", and the "stable" version of bids-validator


sappelhoff:
The run works as expected, see: https://github.com/sappelhoff/bids-examples/runs/1352273053?check_suite_focus=true

however, the output does not seem to be colored (warnings yellow, errors red, pass green) :thinking: any clue why @rwblair ?

sappelhoff:
nice, that seems to work! https://github.com/sappelhoff/bids-examples/runs/1353324397?check_suite_focus=true

rwblair (approved):
Looks good to me, good work!

sappelhoff:
Cool, let's get it in then! thanks for your help in getting the colors work, Ross.

PR #213, opened 2020-11-10
stop validating derivatives dataset until it works

sappelhoff:
This adds a `.SKIP_VALIDATION` file to the derivatives dataset, which will prevent the CI from evaluating it.

Once we want to reiterate on making the validator work for derivatives, a PR can be opened, the first commit being the remove of the `.SKIP_VALIDATION` file - then validation will work in that PR.

### DIFF
--- a/.github/workflows/validate_datasets.yml
+++ b/.github/workflows/validate_datasets.yml
@@ -74,7 +74,7 @@ jobs:
         do
           echo "Validating dataset" $i
           if [ -f ${i%%/}/.SKIP_VALIATION ]; then
-            echo "skipping validation for" ${i%%/}
+            echo "skipping validation for ${i%%/}"
           elif [ -f ${i%%/}/.bids-validator-config.json ]; then
             bids-validator ${i%%/} --ignoreNiftiHeaders || rc=$?
           else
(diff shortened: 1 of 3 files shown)


.github/workflows/validate_datasets.yml
```diff
@@ -73,7 +73,9 @@ jobs:
         for i in $(ls -d */ | grep -v node_modules);
         do
           echo "Validating dataset" $i
-          if [ -f ${i%%/}/.bids-validator-config.json ]; then
+          if [ -f ${i%%/}/.SKIP_VALIDATION ]; then
+            echo "skipping validation for ${i%%/}"
```
effigies, line 77:
```suggestion
            echo "skipping validation for ${i%%/}"; rc=0
```

ds000001-fmriprep/.bidsignore
```diff
@@ -8,3 +8,4 @@ figures/
 *_mixing.tsv
 *_AROMAnoiseICs.csv
 *_timeseries.tsv
+SKIP_VALIDATION
```
effigies, line 11: I don't think this is necessary, since we're not valiidating while it exists.

.github/workflows/validate_datasets.yml
sappelhoff: It seems to work as expected without `rc=0` --> the following datasets continue to be evaluated, but ds000001-fmriprep is properly skipped and for that ds, only an echoed str is printed.

Issue #215, opened 2020-12-03
What is the coordinate system for these datasets (see list)?


sappelhoff:
The following datasets currently all fail validation due to wrong coordinate system naming:

- [x] [ds000117](https://github.com/bids-standard/bids-examples/tree/master/ds000117)
    - [x] (and the "copied" over [eeg_ds000117](https://github.com/bids-standard/bids-examples/tree/master/eeg_ds000117)) --> potentially change `"T1w"` to `"Image"`, depending on how https://github.com/bids-standard/bids-specification/pull/700 goes
- [x] [ds000246](https://github.com/bids-standard/bids-examples/tree/master/ds000246)
- [x] [ds000247](https://github.com/bids-standard/bids-examples/tree/master/ds000247)
- [x]  [ieeg_epilepsy](https://github.com/bids-standard/bids-examples/tree/master/ieeg_epilepsy)
- [x]  [ieeg_epilepsy_ecog](https://github.com/bids-standard/bids-examples/tree/master/ieeg_epilepsy_ecog)
- [x]  [ieeg_filtered_speech](https://github.com/bids-standard/bids-examples/tree/master/ieeg_filtered_speech) --> https://github.com/bids-standard/bids-examples/pull/221
- [x]  [ieeg_motorMiller2007](https://github.com/bids-standard/bids-examples/tree/master/ieeg_motorMiller2007)
- [x]  [ieeg_visual_multimodal](https://github.com/bids-standard/bids-examples/tree/master/ieeg_visual_multimodal)

we should fix them.

As a reminder, the [coordinate systems appendix](https://bids-specification.readthedocs.io/en/latest/99-appendices/08-coordinate-systems.html) in the BIDS specification describes the permissible values for the `*CoordinateSystem*` fields.

Until recently, we did not validate this field properly, so many BIDS datasets have erroneous entries there. The lack of "errors" has also made people believe that all sorts of values can be used for `*CoordinateSystem*` fields, so I imagine that this may come as a surprise for some.

We have several options:
- some datasets may be in an accepted coordinate system, but the value in the JSON field is just wrong --> these will be easy to correct
- other datasets may be in an as-of-yet unaccepted coordinate system, here we have two options:
    - either we expand the "coordinate systems appendix" to turn the unaccepted into and accepted system
    - or we convert the coordinate system of the data to something that is accepted

[...]

PR #219, opened 2020-12-05
ieeg_epilepsy / ieeg_epilepsy_ecog: Fix CoordinateSystem

ftadel:
related #215


sappelhoff (commented):
I seems like I made the impression that I was very strict and determined to have all keywords relating to the [table from appendix VIII](https://bids-specification.readthedocs.io/en/latest/99-appendices/08-coordinate-systems.html#image-based-coordinate-systems) removed --> sorry for that.

I think this does not need to be fixed - instead I will open 2 PRs: one to BIDS-specification (see the plan discussion: https://github.com/bids-standard/bids-specification/issues/661#issuecomment-739800658) and one to the bids-validator that will make coordinate system definitions like the one here valid.

Once all of that is thoroughly discussed and implemented, we can revisit this dataset.

sappelhoff:
From your suite of PRs this is the only one that I can't merge right now. I am hopeful that with https://github.com/bids-standard/bids-specification/pull/700 the change

```diff
- "iEEGCoordinateSystem": "MNI152",
+ "iEEGCoordinateSystem": "Other",
```

won't be necessary and we can keep MNI152, as it's also listed in the table here: https://bids-specification.readthedocs.io/en/stable/99-appendices/08-coordinate-systems.html#standard-template-identifiers

one change that'd have to be made though (if my PR goes through as it is currently) concerns the `sub-ecog01_ses-postimp_space-other_coordsystem.json ` file, where the electrode coords are to be interpreted with regards to `sub-ecog01/ses-postimp/anat/sub-ecog01_ses-postimp_T1w.nii.gz`

there, instead of `"Other"` the keyword `"Image"` would be used if it goes according to the current state of the PR.

sappelhoff:
I am going ahead and will incorporate parts of your PR into #226.

Thanks!

PR #221, opened 2021-01-14
fix filtered_speech ds

sappelhoff:
@choldgraf I made a few edits to your dataset:

1. corrected the "pixels" coordinate system to "Pixels" (capitalization is properly validated now)
2. updated BIDSVersion in dataset_description and added some more metadata I knew to be correct
3. prettified all the JSON to improve human readability
4. changed milliV to mV in accordance with CMIXF-12 rules for unit formatting

unless I hear your objection I would merge these changes.

As far as I know this dataset is not publicly shared, correct?


choldgraf:
go for it 👍

sappelhoff:
Thanks for having a look :-)

PR #226, opened 2021-02-01
Fix remaining coordsystem issues

### LINKED ISSUE #215: What is the coordinate system for these datasets (see list)?

sappelhoff:
closes #219 
closes #215


sappelhoff:
this passes on `master` of bids-validator, so valid to merge and wait for the next validator release.

@ftadel please note how I changed `MNI152` to `MNI152NLin6Sym` in your datasets. Also, `space-other` is changed to `space-Other`.

ftadel:
> @ftadel please note how I changed MNI152 to MNI152NLin6Sym in your datasets. 

@sappelhoff  Thanks for the notice.
However, since this MNI normalization was done using SPM12, according to https://bids-specification.readthedocs.io/en/latest/99-appendices/08-coordinate-systems.html we should probably use **IXI549Space** instead, right?

I'm sorry for this confusion. I followed initially the information from the Lead-DBS MNI page, which states that "MNI152 linear: [...] is used by SPM from version 99 on.":
https://www.lead-dbs.org/about-the-mni-spaces/

@andreashorn Could you please clarify the descriptions of the SPM space on the lead-dbs page? - Thanks

sappelhoff:
I see @ftadel, sorry for the confusion. I opened https://github.com/bids-standard/bids-examples/issues/230 to track this problem. Could you perhaps open a PR to fix the coordinate systems to what they should actually be?

Once you clarified, that is.

andreashorn:
Hi all, 

the info on my post was based on Matt Bretts paper (1) and is likely correct for SPM99 (but not for newer versions).
I love the info page you guys set up that goes way beyond my little post (that became more popular than thought).
I updated it a bit and also referred to your info page prominently from my post.
Hope this helps. So yes, SPM12 uses their own IXI TPM maps and FSL uses the Asym 6th Gen (which is different from the official Sym version published here: http://nist.mni.mcgill.ca/?p=858).

Keep up the good work!

Best, Andy

(1) Brett, M., Johnsrude, I.S., Owen, A.M., 2002. The problem of functional localization in the human brain. Nat Rev Neurosci 3, 243–249. doi:10.1038/nrn756

ftadel:
@andreashorn Thanks for the additional details and for fixing this great reference page! (I also used a lot your pages on all the parcellations/atlases available). Can I ask you for a few more details?

> SPM: [...] This space, referred to as the IXI549Space is largely similar to **MNI152** and of similar resolution 

Do you mean:
- **MNI152Lin**, since this comment appears under that bullet point, or
- **MNI152NLin6Sym**, as according to the BIDS reference page, was used by SPM99 - SPM8 ?

(you can see above that using only `MNI152` was the cause of the inconsistencies this PR tried to address, I guess we should ban it completely from all our communications :)

> Along the same lines, both spaces are unsuitable for use in DBS imaging, in our opinion.

Do you say this only because the spatial resolution is not high enough? 
(1x1x1mm voxel instead of 0.5x0.5x0.5mm for the 09b)

andreashorn:
Hi, I meant to MNI152NLin6Sym – but imho the TPMs in SPM are not that detailed e.g. in cortex and roughly correspond to MNI152 in my view..

Re DBS: It's both resolution and the fact that there's only T1 available and/or only TPMs available (in which for instance, the prime target for DBS, the subthalamic nucleus) is not visible.
But the DBS world is a different world in its own I would say, many things that apply here do not really apply to the fMRI world & vice versa.

PR #233, opened 2021-02-04
ieeg_epilepsy_ecog: Fix FreeSurfer derivatives structure

ftadel:
According to recommendations:
https://surfer.nmr.mgh.harvard.edu/fswiki/BIDS


sappelhoff (commented):
Thanks @ftadel - I have very limited freesurfer experience, so I can't judge whether this makes sense or not --> **however** I *do* see, that the BIDS wiki for freesurfer was last updated some time in 2018. That seems suspicious to me.

Do you know the people responsible for that page and could you perhaps request an update, or that the authors of the page add some note that this page is outdated (or *not* outdated, I cannot judge)

ftadel:
@ahoopes As the author of this page on the FsWiki, could you help us here?

ahoopes:
Sure, responded at https://github.com/bids-standard/bids-specification/issues/461#issuecomment-774225300.

ftadel:
@ahoopes Thank you for the thorough answer!

@sappelhoff @guiomar Shall we proceed to change all the freesurfer paths in the bids-examples then?

sappelhoff:
> Shall we proceed to change all the freesurfer paths in the bids-examples then?

it sounds like both ways are somewhat acceptable? I have no personal preference but if you (Francois) want to change the datasets that you have access to (the full data), then we can gladly mirror the changes on bids-examples.

ftadel:
> it sounds like both ways are somewhat acceptable? 

Yes, both are valid structures, but `sub-ecog01_ses-preimp` is easier to process with FreeSurfer, as the BIDS `/derivatives/freesurfer/` would be directly equivalent to the FreeSurfer `SUBJECTS_DIR`, avoiding any data duplication or complicated-to-maintain symbolic linking. It requires less work on all sides, so I think it's better to make it a clear recommendation.

sappelhoff:
>  so I think it's better to make it a clear recommendation.

okay cool, I think the best way for that would be to state that in the BIDS page of the freesurfer wiki (i.e., streamline/update that page a bit), and then link to the wiki from strategic points in the bids spec.

And of course to update the examples and their associated data.

cc @effigies

ftadel:
Could we merge this modification?
It's been modified in our dataset a while ago.
Thanks!

PR #237, opened 2021-02-04
ieeg_epilepsy: Fix BrainVISA derivatives folder

ftadel:
To match FreeSurfer recommendations:
https://surfer.nmr.mgh.harvard.edu/fswiki/BIDS


sappelhoff:
please see my related comment here: https://github.com/bids-standard/bids-examples/pull/233#pullrequestreview-584104682

ftadel:
@denisri @ylep @dgeffroy @sapetnioc @gauzias
Could you please share your recommendations?

denisri:
Well, nobody has been working yet on making Brainvisa compatible with BIDS, and I'm not the person with the best knowledge of BIDS. That's something we should do anyway...
Schema 1 will cannot be understood by the current organization of Brainvisa, as it stores its data as `<subject>/<modality>/<acquisition>` (`acquisition` being similar to `session` in BIDS, I guess).
Schema 2 could work, using `sub-<subject>_ses-<session>` as subjects names, and this is also how Freesurfer longitudinal does, as I understand, so for a short-term solution it would maybe be better. On the long run, a files hierarchy alternative needs to be written for brainvisa to follow BIDS standards, where I personally find clearer to use something more related to schema 1, but it's not really mandatory (especially if it involves reorganizing lots of existing data files).

ftadel:
Could we please merge this?
(similarly to https://github.com/bids-standard/bids-examples/pull/233)

PR #238, opened 2021-02-04
ieeg_epilepsy: Fix coordinate systems

ftadel:


sappelhoff:
Thanks for digging into this @ftadel

PR #243, opened 2021-02-05
Updated the readme for new users

VisLab:
Once again.....

Here is an initial proposed README (code not tested). My goal was to make it easier for people new to BIDS to try things out. The easiest way is for people to use the bids-validator on the web --- but that isn't yet possible because the interface has no checkbox for ignoring empty raw files.

This is just a start --- improvements are needed/welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,27 +2,57 @@
 
 # bids-examples
 
-This repository contains a set of [BIDS-compatible](https://bids.neuroimaging.io/)
-datasets with **empty raw data files**. These datasets can be useful to:
+This repository contains a collection of [BIDS-compatible](https://bids.neuroimaging.io/)
+datasets with **empty raw data files**. These datasets have been contributed as
+test data for lightweight software tests and to provide examples of how BIDS datasets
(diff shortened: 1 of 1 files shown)


sappelhoff:
I think a new section on https://www.github.com/bids-standard/bids-examples/tree/master/CONTRIBUTING.md would be better suited. Perhaps with a note prominent Link to that file from the readme

VisLab:
Well..... maybe..... I guess that isn't clear to me.

Consider the following use case (which I think will actually be a majority of your users)......

I'm new to bids.... I don't care about "contributing" to bids.... I have to get my data into bids format to deposit somewhere like openNeuro because my grant or the journal I am trying to get published in requires it.  I don't have a "world view" of the BIDS organization.  (Most likely I don't have a clue --- it is easy to forget what a steep hill it is to get started in bids). The first thing I do is look for some examples to run through the validator so I can see what is required.....or even more likely --- I tell a student to go look for some examples to run and follow along.  The starter kit and the examples go hand in hand.  I would like to see our group start to review the examples and really try to clean them up.

PR #268, opened 2021-04-05
participant.tsv participant id updates

rwblair:
update participant ids for genetics_ukbb and filtered speech to pass latest changes to validator


sappelhoff (changes requested):
weird, I thought I did that in https://github.com/bids-standard/bids-examples/pull/266/files ? No time to look into it right now, but it seems like something went wrong

sappelhoff:
The files are already fixed in `master`, see:

https://github.com/bids-standard/bids-examples/blob/master/genetics_ukbb/participants.tsv#L1

https://github.com/bids-standard/bids-examples/blob/master/ieeg_filtered_speech/participants.tsv#L1

I think there is some issue with the GitHub diff here. It shows changes (your "files changes") that shouldn't be changes :thinking: 

perhaps we can rebase this PR on `master` and that will reveal the problem? Or we can just close it. @rwblair

rwblair:
@sappelhoff Good catch! Master should be good.

PR #269, opened 2021-04-13
[FIX] Update PET datasets from "unit" to "units"

mnoergaard:
This PR updates the five PET example datasets to contain fields with "Units" instead of "Unit".


mnoergaard:
Great - thanks, @sappelhoff !

PR #270, opened 2021-04-13
update node in CI

sappelhoff:


sappelhoff:
closing, because CIs pass on master and this PR has lost its point :)

PR #275, opened 2021-06-02
Update README asl_002

patsycle:


sappelhoff:
Thanks for these updates @patsycle !

PR #280, opened 2021-06-07
Update pet002 so it is in accordance with ds001420

mnoergaard:
This PR updates pet002 in the bids-examples so it is in accordance with ds001420 (with empty image files, https://openneuro.org/datasets/ds001420/versions/1.0.1). 

This includes the addition of an additional subject (now two subjects) and session (now baseline and rescan). 

The issue was mentioned here: https://github.com/openneuropet/bids-examples/issues/1#issue-911538823


sappelhoff (approved):
nice, thanks for mirroring the example with the actual data.

PR #284, opened 2021-08-17
Updated eeg_hed_small to be HED-3G compliant

VisLab:
This example is now HED-3G compliant.  Bids validator version >= 1.8.2 is required.

### DIFF
--- a/eeg_hed_small/dataset_description.json
+++ b/eeg_hed_small/dataset_description.json
@@ -1,6 +1,6 @@
 {
     "Name": "Face processing MEEG dataset with HED annotation",
-    "BIDSVersion": "v1.8.2",
+    "BIDSVersion": "v1.6.0",
     "HEDVersion": "8.0.0",
     "License": "CC0",
     "Authors": [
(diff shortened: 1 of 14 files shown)


eeg_hed_small/dataset_description.json
```diff
@@ -1,7 +1,7 @@
 {
     "Name": "Face processing MEEG dataset with HED annotation",
-    "BIDSVersion": "v1.4.0",
-    "HEDVersion": "8.0.0-alpha.2",
+    "BIDSVersion": "v1.8.2",
```
sappelhoff, line 3: The field `BIDSVersion` is not for the bids-validator version, it's for the version of the BIDS specification, which is currently 1.6 --> https://bids-specification.readthedocs.io/en/stable/03-modality-agnostic-files.html#dataset_descriptionjson

the bids-validator version and bids-specification version do not correspond

sappelhoff (commented):
other than my small comment, looks good to me

eeg_hed_small/dataset_description.json
VisLab: Thanks for spotting that.  I've changed to v1.6.0

sappelhoff (approved):

PR #288, opened 2021-09-23
FIX: Remove task entity from ds000117 coordsystems

effigies:
This entity is not indicated by the spec. Either the spec should be amended or the examples fixed, as schema-based validation will mark this as an error.


effigies:
Updated on OpenNeuro. This can be merged here.

sappelhoff (approved):

Issue #244, opened 2021-02-07
task entity for electrodes in eeg_face13?


Remi-Gau:
Starting to use the bids-schema (thanks again @tsalo for this!) in bids-matlab to parse datasets. 

By doing so, my code threw a warning for the [`electrode` files](
https://github.com/bids-standard/bids-examples/tree/master/eeg_face13/sub-001/eeg) that have a `task` entity even though they are not supposed to according to the [spec](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/03-electroencephalography.html#electrodes-description-_electrodestsv).

Is that normal?

sappelhoff:
nope, probably a shortcoming of the validator. I think it was decided that electrodes are typically not applied on a task basis (more on a session, or acq basis), thus the task entity is currently "wrong".

Seems like we'd need to fiddle with the validator regexps

sappelhoff:
as discussed in the maintainers meeting we maybe should re-open the discussion on whether allowing/disallowing a `task` entity for electrodes is warranted. --> before working on a validator PR

sappelhoff:
copying the discussion from the maintainers mattermost here, in order not to forget:

# electrodes.tsv and "task" entity

I thought Chris' argument was as follows:

- if you have a session and several tasks in that session, ...
- but for one task in the session, the electrode cap was removed, ...
- then you wouldn't be able to tell from the BIDS folder+file layout, as `task` is disallowed.

Some situation like:

```
sub-01/
    ses-01/
        eeg/
            sub-01_ses-01_electrodes.tsv
            sub-01_ses-01_coordsystem.json
            sub-01_ses-01_task-nback_eeg.edf
            sub-01_ses-01_task-rest_eeg.edf  <-- electrode cap was removed
            ...     
```

In this example, software would think that `sub-01_ses-01_task-rest_eeg.edf` has the electrode positions from `sub-01_ses-01_electrodes.tsv`, when in fact, it doesn't.

With a `task` entity, this could be solved easily.

--------

--> that's how far I understood the point. 

I might have misunderstood (please correct me).

Having thought about this 2 minutes longer, I think this is not an issue and we don't need to allow `task`.

In the above case, users should just organize as such:

```
sub-01/
    ses-1elec/
        eeg/
            sub-01_ses-1elec_electrodes.tsv
            sub-01_ses-1elec_coordsystem.json
            sub-01_ses-1elec_task-nback_eeg.edf
            ...
    ses-1noelec/
[...]

PR #289, opened 2021-09-23
FIX: Remove task entity from eeg_face13 coordsystems

effigies:
This entity is not indicated by the spec. Either the spec should be amended or the examples fixed, as schema-based validation will mark this as an error.


sappelhoff:
see also #244

sappelhoff:
tagging @andesha: Do you have mirrors of this BIDS example somewhere? If so, do you agree with the proposed changes here and could sync them back to your mirrors?

Andesha:
hello! I have tried to tag a colleague to this post who is more in control of these examples. I will contact them directly and see what they say.

SaraStephenson:
Hello, @Andesha and I have reviewed this and we do agree with this change. We have mirrors and I will sync the proposed changes back to them. We will update our internal scripts that create these file names. 

Should we also do a pull request to remove the task entity in the naming of the electrodes.tsv files as per the specification and conversation in issue [#244](https://github.com/bids-standard/bids-examples/issues/244)?

effigies:
I'm happy to remove task from electrodes as well in this PR, if that's the consensus.

Andesha:
What is the official stance of the standard?

I've done some looking around and there does not seem to be a "rule of law" yet.

effigies:
The standard does not list `task` as a valid entity for electrodes.tsv: [Electrodes description (*_electrodes.tsv)](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/03-electroencephalography.html#electrodes-description-_electrodestsv)

Andesha:
Ah excellent!

Can you remove it as you previously stated? It would be much appreciated.

PR #290, opened 2021-09-23
FIX: Various fixes for ieeg_filtered_speech example

effigies:
* No `space-<label>` entity allowed for `_photo.jpg`
* No `task-<label>` entity allowed for `_electrodes.tsv`
* No `space-<label>` entity allowed for `_coordsystem.json`, if no data files contain `space-<label>`.
* Fixed `IntendedFor` to point to existing files.
   * Note that the subject appears to change, since GitHub is associating a file deleted from `sub-cm8` to one added to `sub-cm4`, and vice-versa.


sappelhoff (approved):
Thanks Chris M. I think these changes can be merged without further ado, because Chris H (the original curator) does not have a "public source" of this BIDS example that would need to be updated (mirroring the changes here).

effigies:
Oh, we should probably cc @choldgraf before merging...

choldgraf:
Sounds good and thanks for the cc.

I don't have time to maintain the examples here, so please feel free to update these as you wish without pinging me!

PR #291, opened 2021-09-28
update global install of bids-validator dev version

sappelhoff:
following https://github.com/bids-standard/bids-validator/pull/1349

### DIFF
--- a/.github/workflows/validate_datasets.yml
+++ b/.github/workflows/validate_datasets.yml
@@ -33,7 +33,7 @@ jobs:
     - uses: actions/checkout@v2
 
     - name: Set up Node.js
-      uses: actions/setup-node@v1
+      uses: actions/setup-node@v2
       with:
         node-version: 12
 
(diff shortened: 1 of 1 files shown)


sappelhoff:
@nellh this flow failed until I upgraded node from 12 to 14 ... and locally I have node 12, and it fails there :thinking: Also, on mne-bids, we also use node 14 (and it works): https://github.com/mne-tools/mne-bids/blob/0fd28f456027104dff94426770d9bef8e96dc713/.github/workflows/unit_tests.yml#L159-L175

Is node 12 vs. 14 an issue here and the new docs we just merged in bids-validator should be updated?

nellh:
> @nellh this flow failed until I upgraded node from 12 to 14 ... and locally I have node 12, and it fails there thinking Also, on mne-bids, we also use node 14 (and it works): https://github.com/mne-tools/mne-bids/blob/0fd28f456027104dff94426770d9bef8e96dc713/.github/workflows/unit_tests.yml#L159-L175
> 
> Is node 12 vs. 14 an issue here and the new docs we just merged in bids-validator should be updated?

The development build does require Node 14 now, but the user requirement is Node 12. The Node 12 end of life is pretty soon, so we could up the requirement to Node 14 for everything.

PR #302, opened 2022-01-13
Updated examples as requested from specification PR #970

VisLab:
This PR is done in conjunction with a rewriting of the `03 HED.md` appendix in the BIDS specification
which corresponds to PR: [https://github.com/bids-standard/bids-specification/pull/970](https://github.com/bids-standard/bids-specification/pull/970).

The following changes were made:

1.  The `eeg_hed_small` was renamed `eeg_ds003654s_hed` and updated to better correspond to the version of `ds003654` on openNeuro.
2.  Additional copies `eeg_ds003654s_hed_inheritance` and `eeg_ds003654s_hed_longform` were added. These datasets are modifications of `eeg_ds003654s_hed` designed to demonstrate different forms of tag organization for testing and illustration.

Review requested: @sappelhoff @dorahermes @Remi-Gau


sappelhoff (commented):
+1, it's good to have these examples (both for testing, and to point users to them). I think now the only thing we are missing is an HED example where an `HED` column in `events.tsv` is used, right? I know that feature is not really how HED should mainly be used ... but if it's a feature, it might as well be tested.

Thanks for editing the README, it's all very clear to me.

You need to remove the `.idea/` folder that you accidentally (?) commited.

CI failures seem to be due to microscopy, which is not yet in "stable" validator

VisLab:
Thanks for spotting the .idea.  Yes, the failures are due to microscopy and not these changes.

We'll add an example of the `HED` column in a future PR. There are a couple of other test cases regarding definition expansion that will also need to be added in a future PR. 

I'd like to get this PR through with the `HED.md` update PR, so that we can move forward on the library schema update.

sappelhoff (approved):

sappelhoff:
Merged- thanks @VisLab

PR #304, opened 2022-01-23
Move some examples to use BIDS URIs

### LINKED ISSUE #303: Edit examples to use BIDS URIs
- https://github.com/bids-standard/bids-specification/pull/918

... will introduce BIDS URIs, and we should have some coverage for this in the examples, also to test the validator, see:

- https://github.com/bids-standard/bids-validator/pull/1404
- https://github.com/bids-standard/bids-validator/pull/1409

The PR will also deprecated `RawSources`, and recommend to use `Sources` instead, with a BIDS URI pointing to the raw dataset.

This is convenient, because we could do a single PR, replacing several of the uses of `RawSources`: Removing deprecated fields, and replacing them with BIDS URIs that we want to display/test.

Use `grep -r "RawSources"` to identify candidates. For example:

`qmri_mp2rage/derivatives/pymp2rage/sub-1/anat/sub-1_UNIT1.json`

```
   1   │ {
   2   │   "RawSources": [
   3   │     "sub-1/anat/sub-1_inv-1_part-mag_MP2RAGE.nii",
   4   │     "sub-1/anat/sub-1_inv-1_part-phase_MP2RAGE.nii",
   5   │     "sub-1/anat/sub-1_inv-2_part-mag_MP2RAGE.nii",
   6   │     "sub-1/anat/sub-1_inv-2_part-phase_MP2RAGE.nii"
   7   │   ],
   8   │   "EstimationReference": "Marques et al., 2010",
   9   │   "EstimationAlgorithm": "MP2RAGE T1UNI",
  10   │   "EstimationSoftwareName": "pymp2rage",
  11   │   "EstimationSoftwareVer": "v0.1",
  12   │   "EstimationSoftwareLang": "Python 3.6",
  13   │   "EstimationSoftwareEnv": "Mac OS X 10.14.5"
  14   │ }
```
Replace with:

```
"Sources": [
    "bids:raw:sub-1/anat/sub-1_inv-1_part-mag_MP2RAGE.nii",
    ...
]
```

and add a `dataset_description.DatasetLinks` object with a key `raw`.

EDIT: Potential problem:
- when the `DatasetLinks` is added to a `dataset_description` nested in `derivatives/`, then it won't be validated
- this is fine, BUT
- we should think about additional examples that we could modify to fully display the new BIDS URI feature
- this also includes modifying some `IntendedFor` fields to use a syntax like `bids::sub-01/fmap/sub-01_dir-AP_epi.nii.gz` (meaning: relative path within a local dataset)

sappelhoff:
Closes #303 

I don't think we need to move all examples to use BIDS URIs (at least not all at once), what do you think? The old behavior stays valid (backward compatibility) after all.

In this PR, I touched 3 datasets:
- `7t_trt`, because it was easy to do
- `ieeg_filtered_speech`, because no "full data" exists for this dataset, so we can change it without having to mirror something back
- `ds000117`, ... here I started and stopped again, because a lot would need to be done, AND this dataset actually exists and is widely used, so we perhaps need to be more careful here.

This is what I changed (using regex search+replacement in VScode, limiting the search to the respective dataset via the "include" option):

#### 7t_trt

- search: `"IntendedFor": "ses-(\d)/func/sub-(\d\d)_ses-\d_task-rest_acq-fullbrain_run-(\d)_bold.nii.gz"`
- replace: `"IntendedFor": "bids::sub-$2/ses-$1/func/sub-$2_ses-$1_task-rest_acq-fullbrain_run-$3_bold.nii.gz"`
- update `dataset_description.json`

#### ieeg_filtered_speech

- search: `"IntendedFor": "/sub-(..\d+)/ieeg/sub-..\d+_photo.jpg"`
- replace: `"IntendedFor": "bids::sub-$1/ieeg/sub-$1_photo.jpg"`
- update `dataset_description.json`

#### ds000117

- fix fake arrays in emptyroom files `"a; b; c"` --> `["a", "b", "c"]`

#### ds000246

- manually replace `AssociatedEmptyRoom` fields with local BIDS URIs (=simply prepending `bids::` to the already-dataset-relative links and removing the leading slash after the `::`)

#### synthetic

replace `RawSources` with BIDS URIs via a `DatasetLinks` object in the derivatives `dataset_description.json`, using:

- search: `    "(sub-.+/ses-.+/sub-.+_ses-.+_task-.*_bold.nii")`
- replace: `    "bids:raw:$1`
- search: `"RawSources": \[`
- replace: `"Sources": [`

---
I also bumped the BIDSVersion of these datasets to 1.8.0, which will be the version that supports bids uris.

---

- [x] I think what would be good to do as well is to "deprecate" a few of the `RawSources` fields that we have in some examples, as mentioned in

- https://github.com/bids-standard/bids-examples/issues/303#issue-1105064298

~~Waiting for an answer on this thread to get started with that:~~

- https://github.com/bids-standard/bids-specification/pull/918#discussion_r790291925

- [x] Furthermore we should have the `AssociatedEmptyRoom` field in BIDS URIs in at least one dataset.
- [x] And have a `DatasetLinks` object in at least one `dataset_description.json`
    - Note that unfortunately this is only in a derivatives `dataset_description.json`, and derivatives are currently not validated. However I did not see any other dataset where I could have added this to the "raw" dataset description. So I think this is as good as it gets for now.

What else should we do before we can feel comfortable merging BIDS URIs into the BIDS specification? Suggestions?

### DIFF
--- a/ds000246/sub-0001/meg/sub-0001_task-AEF_run-01_meg.json
+++ b/ds000246/sub-0001/meg/sub-0001_task-AEF_run-01_meg.json
@@ -23,4 +23,4 @@
 "MaxMovement":0.0009892724421254463,
 "DigitizedLandmarks":true,
 "DigitizedHeadPoints":true,
-"AssociatedEmptyRoom":"bids::/sub-emptyroom/meg/sub-emptyroom_task-noise_run-01_meg.ds"}
+"AssociatedEmptyRoom":"bids::sub-emptyroom/meg/sub-emptyroom_task-noise_run-01_meg.ds"}
--- a/ds000246/sub-0001/meg/sub-0001_task-AEF_run-01_meg.json
+++ b/ds000246/sub-0001/meg/sub-0001_task-AEF_run-01_meg.json
@@ -23,4 +23,4 @@
 "MaxMovement":0.0009892724421254463,
 "DigitizedLandmarks":true,
 "DigitizedHeadPoints":true,
-"AssociatedEmptyRoom":"/sub-emptyroom/meg/sub-emptyroom_task-noise_run-01_meg.ds"}
+"AssociatedEmptyRoom":"bids::/sub-emptyroom/meg/sub-emptyroom_task-noise_run-01_meg.ds"}
--- a/ds000246/sub-0001/meg/sub-0001_task-AEF_run-02_meg.json
+++ b/ds000246/sub-0001/meg/sub-0001_task-AEF_run-02_meg.json
@@ -24,4 +24,4 @@
 "MaxMovement":0.0032890058287054712,
 "DigitizedLandmarks":true,
 "DigitizedHeadPoints":true,
-"AssociatedEmptyRoom":"/sub-emptyroom/meg/sub-emptyroom_task-noise_run-01_meg.ds"}
+"AssociatedEmptyRoom":"bids::/sub-emptyroom/meg/sub-emptyroom_task-noise_run-01_meg.ds"}
--- a/ds000246/sub-0001/meg/sub-0001_task-AEF_run-02_meg.json
+++ b/ds000246/sub-0001/meg/sub-0001_task-AEF_run-02_meg.json
@@ -24,4 +24,4 @@
 "MaxMovement":0.0032890058287054712,
 "DigitizedLandmarks":true,
 "DigitizedHeadPoints":true,
-"AssociatedEmptyRoom":"bids::/sub-emptyroom/meg/sub-emptyroom_task-noise_run-01_meg.ds"}
+"AssociatedEmptyRoom":"bids::sub-emptyroom/meg/sub-emptyroom_task-noise_run-01_meg.ds"}
--- a/.github/workflows/validate_datasets.yml
+++ b/.github/workflows/validate_datasets.yml
@@ -48,7 +48,7 @@ jobs:
         pushd ..
         # Get npm 7+
         npm install -g npm@^7
-        git clone --depth 1 https://github.com/bids-standard/bids-validator
+        git clone --depth 1 https://github.com/sappelhoff/bids-validator/ --branch intendedfor/bidsuri
         cd bids-validator
         # Generate the full development node_modules
         npm clean-install
--- a/.github/workflows/validate_datasets.yml
+++ b/.github/workflows/validate_datasets.yml
@@ -48,7 +48,7 @@ jobs:
         pushd ..
         # Get npm 7+
         npm install -g npm@^7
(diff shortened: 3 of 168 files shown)


sappelhoff:
For an explanation of current CI failure, please see https://github.com/bids-standard/bids-validator/issues/1393#issuecomment-1019520428

ds000246/sub-0001/meg/sub-0001_task-AEF_run-01_meg.json
```diff
@@ -23,4 +23,4 @@
 "MaxMovement":0.0009892724421254463,
 "DigitizedLandmarks":true,
 "DigitizedHeadPoints":true,
-"AssociatedEmptyRoom":"/sub-emptyroom/meg/sub-emptyroom_task-noise_run-01_meg.ds"}
\ No newline at end of file
+"AssociatedEmptyRoom":"bids::/sub-emptyroom/meg/sub-emptyroom_task-noise_run-01_meg.ds"}
```
effigies, line 26: We dropped the initial `/`, right?

```suggestion
"AssociatedEmptyRoom":"bids::sub-emptyroom/meg/sub-emptyroom_task-noise_run-01_meg.ds"}
```

ds000246/sub-0001/meg/sub-0001_task-AEF_run-02_meg.json
```diff
@@ -24,4 +24,4 @@
 "MaxMovement":0.0032890058287054712,
 "DigitizedLandmarks":true,
 "DigitizedHeadPoints":true,
-"AssociatedEmptyRoom":"/sub-emptyroom/meg/sub-emptyroom_task-noise_run-01_meg.ds"}
\ No newline at end of file
+"AssociatedEmptyRoom":"bids::/sub-emptyroom/meg/sub-emptyroom_task-noise_run-01_meg.ds"}
```
effigies, line 27:
```suggestion
"AssociatedEmptyRoom":"bids::sub-emptyroom/meg/sub-emptyroom_task-noise_run-01_meg.ds"}
```

ds000246/sub-0001/meg/sub-0001_task-AEF_run-01_meg.json
sappelhoff: good catch, we actually don't talk about leading slashes in the `<relative-path>` component. I have made a suggestion here: https://github.com/bids-standard/bids-specification/pull/918#discussion_r928108876

I think trailing slashes in the `<dataset-name>` component are fine.

effigies (approved):

PR #306, opened 2022-01-30
[FIX] change several qMRI examples `FieldStrength` to `MagneticFieldStrength`

Remi-Gau:
Had not noticed that one.


Remi-Gau:
@Gilles86 I suspect that those should also be fixed on the OSF examples.

sappelhoff (approved):

Remi-Gau:
errors on this seem to be coming from validation of microscopy datasets but only for "stable".

https://github.com/bids-standard/bids-examples/runs/5205366510?check_suite_focus=true#step:8:1959

effigies (approved):

PR #305, opened 2022-01-27
fNIRS example dataset (BEP030)

rob-luke:
Dataset with functional near-infrared spectroscopy measurements in BIDS format.

- BEP030: https://github.com/bids-standard/bids-specification/pull/802
- Validator PR: https://github.com/bids-standard/bids-validator/pull/952

Complete dataset (without zeroed data) can be found at https://github.com/rob-luke/BIDS-NIRS-Tapping

PR #317, opened 2022-06-20
add another fNIRS example with finger/foot tapping during an automaticity task

robertoostenveld:
this is a dataset that will be shared in full by @helenacockx on the [Donders Repository](https://data.donders.ru.nl). 

See also https://github.com/bids-standard/bids-examples/pull/305#issuecomment-1160090888


sappelhoff:
Thanks Robert, but the first two commits seem to have "sneaked in" -- could you update the branch to only contain your commits, please?

robertoostenveld:
Sorry, I don't know how that happened, I though I started with master. I will start a new branch or revert these, and also get rid of the large pdf file.

robertoostenveld:
Oh I see now, I should have started with `bep030_nirs`, not `master`

sappelhoff:
In that case, maybe this problem can also be solved by updating `bep030_nirs`

robertoostenveld:
oops ...
I did not merge anything, I just removed the 4 commits to make my branch aligned with `bep030_fnirs`

sappelhoff:
yeah, I've had this problem before ... GitHub can close your branch when there are "no changes" (even if you intend to just clean up and then push the real work), I suspect that's what happened here.

PR #323, opened 2022-06-20
[ENH] New fNIRS example dataset: (non-)auto finger/foot tapping

robertoostenveld:
this replaces #317 and follows up on [this conversation](https://github.com/bids-standard/bids-examples/pull/305#issuecomment-1160090888)


sappelhoff (commented):
We can ignore the checks of the "stable" validator -- but the "master" checks should succeed.

There is an issue with the phenotype folder that relates to some points in this discussion: https://github.com/bids-standard/bids-specification/issues/914

[see CI check](https://github.com/bids-standard/bids-examples/runs/6966044586?check_suite_focus=true#step:8:1785)

There are also more errors that could be interesting to look into to either fix the validator, this example, or the spec itself :+1: 

Thanks!

sappelhoff:
> There was one issues that I resolved using the .bidsignore file: the phenotype/practicelogbook.tsv has multiple rows for each participant (since they practiced the same task multiple times), but the validator does not seem to like this.

Why not reformat the dataframe to "wide" format, e.g.

```
id ses1 ses2 ses3
1 22 32 11
2 44 34 22
```

instead of 

```
id ses val
1 1 22
1 2 32
1 3 11
2 1 44
2 2 34
2 3 22
```

robertoostenveld:
> > There was one issues that I resolved using the .bidsignore file: the phenotype/practicelogbook.tsv has multiple rows for each participant (since they practiced the same task multiple times), but the validator does not seem to like this.
> 
> Why not reformat the dataframe to "wide" format, e.g.
> 
> ```
> id ses1 ses2 ses3
> 1 22 32 11
> 2 44 34 22
> ```
> 
> instead of
> 
> ```
> id ses val
> 1 1 22
> 1 2 32
> 1 3 11
> 2 1 44
> 2 2 34
> 2 3 22
> ```

Because of human readability reasons: It has 8 columns (i.e. 7 data columns), which would result in a wide format that is 1+5*7=36 columns wide. On top of that, one of the 8 columns (the last) is "notes" in a free-text format that can contains whole sentences (or n/a when not specified).

An alternative solution that we considered (but also decided against) is to split it over 5 files. That also negatively affects the human readability, as you would have to scan over 5 files to see how performance (and subjective experience) changes over the 5 training days.

Since this dataset is not meant as a phenotype example, and since the phenotype validation is apparently not optimal yet, I consider it fine if the phenotype data in this dataset is treated as exception. I could also leave it out, but then the test dataset would diverge from the actual one.

rob-luke:
> With two small changes to the nirs branch of the bids-validator of @rob-luke this dataset now passes the validation

Thank you Robert for this PR. It is now merged. Can someone please kick the CI to see if the master now goes green.

rob-luke:
This is a wonderful example dataset. I learnt lots from looking through it. It provides a great compliment to #305, beyond being a different device it has many differences to 305 including:

* this dataset has x/y/z and mine has template/x/y/z,
* this one has more complex optode names
* this one has different nominal and actual wavelength
* this dataset provides the data in optical density rather than cw amplitude
* this one  short channels specified
* much more detailed metadata and descriptions
* etc

Thanks for adding this example. It all looks great to me

sappelhoff:
> Thank you Robert for this PR. It is now merged. Can someone please kick the CI to see if the master now goes green.

done, and it comes back green :-)

robertoostenveld:
> This is a wonderful example dataset. ...

That compliment goes to @helenacockx

helenacockx:
Thank you, Robert, for the adaptations to make it pass through the validator!
I am currently traveling through the USA and Canada to visit some labs here, so I don't have much time to look closely at it, but I will do it once I find the time!

helenacockx:
The updates from Robert look excellent! I have nothing more to add.

sappelhoff:
Thanks @robertoostenveld

PR #329, opened 2022-08-05
[ADD] Motion (BEP029): example data set "dualtask"

sjeung:
Adding another example data set from a dual task (EEG + motion)


motion_dualtask/dataset_description.json
```diff
@@ -0,0 +1,19 @@
+{
+	"Acknowledgements":"n\/a",
```
sappelhoff, line 2: why are the slashes escaped?

sjeung: It has to do with JSONLab, which I used for creating this dataset. Apparently the official JSON syntax allows it, too, but I had suggested an update to skip escape. Please see [this thread ](https://github.com/fangq/jsonlab/issues/82).

sappelhoff (commented):
this is a HUGE example dataset (3009 files), does it need to be that big, or could you also imagine pruning it down a little? For example by supplying fewer subjectes than originally present in the full dataset?

motion_dualtask/dataset_description.json
sappelhoff: The thread tells me that you found a solution, are you still planning to propagate the solution here?

rwblair:
> this is a HUGE example dataset (3009 files), does it need to be that big, or could you also imagine pruning it down a little? For example by supplying fewer subjectes than originally present in the full dataset?

All the subjects are about the same size, around 1mb or so and have similar number of files (between 75 and 141). So there's no obvious low hanging fruit.

sappelhoff (commented):
@sjeung could you please remove the `._.DS_Store` file from the PR and potentially add it to `.gitignore`?

Other than that, should we merge this into the BEP029 branch now?

sjeung:
> @sjeung could you please remove the `._.DS_Store` file from the PR and potentially add it to `.gitignore`?
> 
> Other than that, should we merge this into the BEP029 branch now?

Yes, it is more or less final as it is now. Thanks!

PR #332, opened 2022-08-28
Hed library example addition

VisLab:
This addition is in support of bids-standard/bids-specification#1106 and bids-standard/bids-validator#1496.


sappelhoff (commented):
do you think it's important that this gets its own example? Or could you imagine editing the already existing `eeg_ds003654s_hed` example?

I think whenever possible, we should try to work with fewer, rather than more examples.

VisLab:
I think for now this needs its own example.  The validation takes a different code path and I think it needs to be tested.  

When the SCORE library 1.0.0 is released, I believe that some of the IEEG examples of @dorahermes @tpatpa will incorporate HED library schema.  I decided to put this in over the weekend because I wanted to be sure that the new code paths were tested.

When SCORE stuff gets in, I'll remove this example.  I also think the longform example can be consolidated with the base example.  I'll consolidate those two examples also when I do the removal. In the meantime, I would appreciate it going in.

sappelhoff:
Alright, makes sense to me. I'll wait for the validator to be merged (probably as soon as Ross has time to have a final look) and then we can re-run the CI checks here and merge if they come back green.

sappelhoff (approved):
Now tests are passing on `master`, as expected :+1:

PR #337, opened 2022-10-21
[REF] refactor qmri metadata to add JSON in root folder

Remi-Gau:
relates to:

- https://github.com/bids-standard/bids-validator/pull/1546

validation failure is expected until PR 1546 of the validator is merged

modifying those datasets should act as a "regresssion test" for the bug:
- https://github.com/bids-standard/bids-validator/issues/1438


Remi-Gau:
pinging @agahkarakuzu to keep him in the loop on this.

qmri_megre/MEGRE.json
```diff
@@ -0,0 +1,8 @@
+{
+
+	"MagneticFieldStrength": "3",
+	"Manufacturer": "Siemens",
+	"ManufacturerModelName": "TrioTim",
+	"PulseSequenceType": "GR",
+	"EchoTime": "0.02"
```
agahkarakuzu, line 7: I think the `EchoTime` should not be in this top level file for `MEGRE` as it changes.

agahkarakuzu:
Hi @Remi-Gau sorry I could get back to this just now. Just to make sure that I understand, the purpose it to collect parameters that do not change across file collection under one top json file that has the name of the suffix right? 

I remember we discussed something like this to remove redundancy, but was not an option at the time we were developing the standard.

qmri_mp2rage/sub-1/anat/sub-1_inv-1_MP2RAGE.json
```diff
@@ -1,8 +1,3 @@
 {
-  "InversionTime": 0.8,
-  "FlipAngle": 5,
```
agahkarakuzu, line 3: We should keep `FlipAngle` for individual MP2RAGE images.

qmri_megre/MEGRE.json
agahkarakuzu, line 7: `EchoTime` changes for MEGRE, so should not be in the common file?

qmri_mtsat/MTS.json
```diff
@@ -0,0 +1,7 @@
+{
+  "MagneticFieldStrength": "3",
+  "Manufacturer": "Siemens",
+  "ManufacturerModelName": "TrioTim",
+  "PulseSequenceType": "SPGR",
+  "RepetitionTime": "0.028"
```
agahkarakuzu, line 6: `RepetitionTime` may change along with FA. It is not the case for this one, but a possibility to keep in mind.

qmri_vfa/VFA.json
```diff
@@ -0,0 +1,7 @@
+{
+  "MagneticFieldStrength": "3",
+  "Manufacturer": "Siemens",
+  "ManufacturerModelName": "TrioTim",
+  "PulseSequenceType": "SPGR",
+  "RepetitionTime": "0.0150"
```
agahkarakuzu, line 6: Similar to MTS, `RepetitionTime` may be different. Do we need to make some notes about this, or is it clear to the users (somewhere in the spec and maybe even at the validation level) that the parameters are listed under common file only when they are constant across file collection pairs?

agahkarakuzu (changes requested):

Remi-Gau:
> Hi @Remi-Gau sorry I could get back to this just now. Just to make sure that I understand, the purpose it to collect parameters that do not change across file collection under one top json file that has the name of the suffix right? 
> 
> I remember we discussed something like this to remove redundancy, but was not an option at the time we were developing the standard. 
> 
> 

Well yes and no.

Because of the Inheritance principle, users can still do that (meaning they can put json files in the root folder that contain metadata shared across files downstream).

But currently if they do so then the validator complains and does not recognized those files as valid.

So I am modifying those datasets to demonstrate the bug and create a regression test.

I have not given much thought on how to allocate metadata in the root folder and just tied to dump all common metadata there. Apparently I made some mistake.

Also I agree that it may be wiser to put only in there the metadata that make the most sense from a qmri perspective given that those datasets are pretty much a "first line of defense" when it comes to "good" small examples.

Remi-Gau:
Another thing. There is a tacit rule that bids examples datasets can diverge from the real datasets they originated from. So my changes here do not have to be reflected in the ones you have on OSF.

qmri_vfa/VFA.json
Remi-Gau: Will cross check. Not sure about the spec or the validator about this.
Best way to quickly know for the validator is to stress test it.

Remi-Gau:
Will adress other changes when I get access to a computer. So in a few days.

agahkarakuzu:
I see, thank you so much Remi!

Remi-Gau:
@agahkarakuzu I updated which metadata are in the root of the datasets.

From my reading of the qmri section, there is nothing that jumps at me that could be unclear regarding how to spread the metadata around.

Remi-Gau:
@sappelhoff if we mostly care about the master branch of the validator passing then this should be good to go as well, no?

sappelhoff:
> @sappelhoff if we mostly care about the master branch of the validator passing then this should be good to go as well, no?

yes, the stable branch will fail until MOTION additions are released on the validator repo.

If this is ready, let's merge it as well 👍 will someone mirror the changes to the real data?

Remi-Gau:
I remember that we had a soft agreement that datasets in the examples can start diverging from their uostream dataset and that we do not have to keep them in synch.

But if @agahkarakuzu feels this would be anice addition to the real examples on osf I could give it a go.

PR #347, opened 2023-02-01
Bep029 motion - Example systemvalidation

JuliusWelzel:
This is a new PR because the old one is not actively maintained anymore.
 The added example follows the newest BEP029 specs and is tested with the #1434 PR from the BIDS-Validator


JuliusWelzel:
@sappelhoff maybe you want to add labels and have a look if everything is correctly done?

JuliusWelzel:
> What do you mean by adding labels?
You usually added labels like `BEP` and I was not sure how to do this :D

> I see that `/` in JSON is still escaped with backslashes, which I think is unnecessary. But @effigies said it doesn't matter so it's fine by me to keep it this way.
Fixed
 
> Other than that, you didn't update: https://github.com/bids-standard/bids-examples/blob/master/README.md
 
> and remember what I wrote in the maintainers biweekly meeting notes doc about changing the [continuous integration script](https://github.com/bids-standard/bids-examples/blob/master/.github/workflows/validate_datasets.yml) so that it can use the MOTION branch of the validator? --> that could also be done.
> 
> Alternatively we can just merge this and you continue to do this work on the `bep029_motion` branch :-)
> 
> let me know
If it is fine for you I am happy to merge on the `bep029_motion` branch

PR #348, opened 2023-02-02
BEP029 Motion

sappelhoff:
This PR is to be merged into the `master` branch when the BEP029 Motion is integrated into the BIDS specification.

Leads:

- @sjeung 
- @JuliusWelzel

PR #354, opened 2023-02-10
Bep029 motion dualtask

sjeung:
Addresses the [issues] (https://github.com/bids-standard/bids-examples/pull/348#pullrequestreview-1289022140) @sappelhoff  pointed out regarding dual task data set.

- Extra json fields removed : there was one participant whose json was not updated 
- README - added

- undefined columns warning seems to not go away when the .tsv files have no headers 
The warning comes from motion.tsv files not channels.tsv and the same goes true for spot rotation data set. 
I tried the version of motion.tsv that actually has (data but not header) entries in columns but the warning persists.


sappelhoff:
> I tried the version of motion.tsv that actually has (data but not header) entries in columns but the warning persists.

okay, that's something we need to fix, can you put it on your to do list for the validator and write an issue on bids-validator? It's not a blocker for progressing, we can fix it anytime soon.

PR #356, opened 2023-02-15
Convert any sample columns in events files to integers.

rwblair:
As per:
https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/05-task-events.html

script I ran from root of the repository:
```
#!/bin/bash
find  . -name "*_events.tsv" -exec grep -l sample {} \; | while read line;
    do indx=$(head -n1 $line | tr "\t" "\n" | grep -nx sample |  cut -d":" -f1);
    if [[ $indx ]]; then 
        if grep -lzP "\r\n$" $line; then
            rs="\r\n"
        else
            rs="\n"
        fi
        echo "$line at column $indx"
        gawk -v RS=$rs -v col=$indx 'BEGIN { FS=OFS="\t" } { $col=gensub(/([0-9]+)\.[0-9]+/, "\\1",1, $col); print }' $line > tmp && mv tmp $line;

    fi
done
```


eeg_face13/sub-006/eeg/sub-006_task-faceFO_events.tsv
```diff
@@ -1,300 +1,300 @@
 onset	duration	trial_type	response_time	sample	value
```
effigies, line 1: Samples with .5

eeg_ds003654s_hed/sub-002/eeg/sub-002_task-FacePerception_run-1_events.tsv
```diff
@@ -1,553 +1,553 @@
-onset	duration	sample	event_type	face_type	rep_status	trial	rep_lag	value	stim_file
-0.004	n/a	1.0	setup_right_sym	n/a	n/a	n/a	n/a	3	n/a
-24.2098181818	n/a	6052.4545	show_face_initial	unfamiliar_face	first_show	1	n/a	13	u032.bmp
```
effigies, line 3: Samples with lots of extra digits

effigies (commented):
It would be good to get feedback from the `eeg_face13` contributor, which has non-integer samples (but halves) and `ds003654s_*` which has the `onset` column divided by `0.004`.

I agree that samples should be integers and this is explicit in the spec.

effigies:
https://github.com/bids-standard/bids-examples/blob/280478ef7f9167300ed06a8b2a9727b108056603/eeg_face13/dataset_description.json#L4-L7

https://github.com/bids-standard/bids-examples/blob/280478ef7f9167300ed06a8b2a9727b108056603/eeg_ds003654s_hed/dataset_description.json#L6-L13

@VisLab you might be a good person to check in with here.

sappelhoff (commented):
thanks Ross. I'll take care of mirroring these changes to `eeg_matchingpennies`, where apparently I used floats instead of integers ... so I'll just cut off the `.0` at each sample.

Some further notifications:

- for `eeg_rishikesh` --> @arnodelorme 
- for `fnirs_tapping` --> @rob-luke 
- for `eeg_face13` --> @Andesha, I believe

christinerogers:
@Andesha is this something we could update in the next month perhaps? 

(thanks @sappelhoff for bringing my attention to this)

sappelhoff:
> I thought I had caught and corrected all of the 54 --> 45 errors. Where did this one come from @effigies? I'd like to correct that while I am at it. Thx

That is simply a git merge conflict. In `master`, the files are named correctly, see e.g.:

- https://github.com/bids-standard/bids-examples/tree/master/eeg_ds003645s_hed

they show up here as wrong because @rwblair's branch is old

Andesha:
Feel free to cut off the decimal values for the samples. That's just a result of down sampling from the original recording of 1024Hz to 512Hz.

effigies:
I reran Ross' script on `master` and pushed to #362.

@VisLab To confirm, truncating to round down to the nearest int is an acceptable transform?

arnodelorme:
Sample latencies can sometimes be fractional because they come from a different machine that has more resolution than the EEG sampling frequency. This is important for reaction time, for example. Even with EEG sampled at 250 Hz, you need to be able to determine reaction time with 1ms (1000 Hz equivalent sampling rate) precision.

effigies:
`sample` was introduced in BIDS 1.2.0, saying only:

> Column name | Description
> -- | --
> sample | OPTIONAL. Onset of the event according to the sampling  scheme of the recorded modality (i.e., referring to the raw data file  that the events.tsv file accompanies).

In 1.7.0, it became:

> Column name | Requirement Level | Data type | Description
> -- | -- | -- | --
> sample | OPTIONAL | integer | Onset of the event according to the sampling scheme of the recorded modality (that is, referring to the raw data file that the events.tsv file accompanies).

The integer type was introduced in https://github.com/bids-standard/bids-specification/commit/034c6a85fcf8eac246951c2ade29cebccbe57775 (part of bids-standard/bids-specification#827) as part of the schematization. There doesn't appear to have been written discussion on whether sample is an integer or a float. From @arnodelorme's post, it sounds like fractional samples are a desired feature, so it's less an index than an alternative unit of time. From this perspective, the fix is not truncation but relaxing the data type to be `number`.

There is an open issue about whether it is a 0- or 1-based index here: https://github.com/bids-standard/bids-specification/issues/499. If it is an alternative temporal unit, then I think starting at 0 more sense than starting at 1.

VisLab:
Chris,
I think it should be rounded.....

VisLab:
One of the ongoing problems with the BIDS onset, is that there is no
precision specification on the onset.  I have seen some EEG datasets
sampled at say 200 Hz which report event onsets rounded to the nearest
int, making them useless in analysis as onsets are in seconds.

Sample has the potential to specify the position of an event relative to
the underlying data a more accurate precision if you know the sampling rate
of the data that the event is tied to.

I agree with Arno that it might make sense to leave it as real numbers (or
doesn't hurt anything as long as it is clear what it means), especially
when the events are generated by a device that is being sampled
independently and at a different rate.

I would agree that if it is an alternative temporal unit it should start at
0.

dorahermes:
+1 on zero-based indexing

effigies:
Thanks for the quick responses, Dora and Kay. I've proposed https://github.com/bids-standard/bids-specification/pull/1441, if any interested parties from this thread would care to comment there.

sappelhoff:
The `sample` column will be considered an arbitrary column in future BIDS version. As such, its use is dataset-specific and SHOULD be documented in the accompanying JSON files. Thanks for the discussion in #1441, everyone.

PR #360, opened 2023-03-02
switch true and false to lower case in all channels.tsv for fnirs_automaticity

rwblair:
Commands run:
`find . -name "*_channels.tsv" -exec sed -ir 's/\tTRUE/\ttrue/g' {} \;`
`find . -name "*_channels.tsv" -exec sed -ir 's/\tFALSE/\tfalse/g' {} \;`


sappelhoff:
cc @robertoostenveld @helenacockx

Issue #367, opened 2023-05-04
Duplicated rows causing index validation error for motion_spotrotation


rwblair:
One example:
https://github.com/bids-standard/bids-examples/blob/master/motion_spotrotation/sub-02/ses-joy/sub-02_ses-joy_scans.tsv

sjeung:
This was just an error in the scans.tsv files and the lines can go

PR #372, opened 2023-05-31
[FIX] remove duplicated scans.tsv rows 

sjeung:
Fixing erroneously duplicated scans.tsv rows

See issue #367


sappelhoff:
Thanks!

Issue #373, opened 2023-06-02
Reduce the size of the bids examples repository


Remi-Gau:
the bids examples repo is now close to a 100 Mo: this is really not ideal for something meant to be lightweight

Most of the weight comes from the stimuli examples.

```bash
$ du -h -d 0  */stimuli | sort -rh
23M    eeg_ds000117/stimuli
8,1M    eeg_ds003645s_hed/stimuli
8,1M    eeg_ds003645s_hed_longform/stimuli
8,1M    eeg_ds003645s_hed_inheritance/stimuli
2,0M    fnirs_automaticity/stimuli
```

Given that sitmuli folder content is almost completely unspecified in the BIDS specification it seems excessive that they would take so much space.

Discussed during the last maintainers meeting, our first approach would be to zero the content of any stimuli file in there and then rewrite the history.

Pinging @VisLab to get the opinion of the HED team.
Should we be concerned that not having examples of the actual stimuli images may reduce the usefulness of the examples when it to understand the hed tags?

Remi-Gau:
What ever decision we agree on, the history of the repo may need to be rewritten:

https://github.com/bids-standard/bids-examples/issues/119

VisLab:
We can remove the stimuli directories.  The dataset on OpenNeuro has them.
Do you want me to do it?

Remi-Gau:
> We can remove the stimuli directories. The dataset on OpenNeuro has them. Do you want me to do it?


If you can do it that'd be much appreciated.

No need to remove them, you can simply truncate the datafiles in them:

https://github.com/bids-standard/bids-examples/blob/master/CONTRIBUTING.md#how-to-truncate-data-files-to-0kb

Remi-Gau:
Amended the title of this issue to reflect the steps left to do

yarikoptic:
FWIW, recently-ish added `git-replace` command could be of help after history rewrite to establish "continuity" of history. See e.g. https://andrewlock.net/reducing-the-size-of-a-git-repository-with-git-replace/ .

PR #376, opened 2023-06-02
truncate files in stimuli directories for some datasets

VisLab:
relates to #373


Remi-Gau:
thanks @VisLab 

@sappelhoff we should also probably zero the stim from  https://github.com/bids-standard/bids-examples/tree/master/ds000117/stimuli

no?

sappelhoff:
> @sappelhoff we should also probably zero the stim from https://github.com/bids-standard/bids-examples/tree/master/ds000117/stimuli

agreed.

sappelhoff:
we reduce around 40-50MB with the changes here, from my side that's good enough for now. However if we will rewrite history later (to actually shrink the repo size for full git clones), then perhaps it'd be good to reduce as much as possible (so yes for reducing automaticity as well then).

Remi-Gau:
> we reduce around 40-50MB with the changes here, from my side that's good enough for now. However if we will rewrite history later (to actually shrink the repo size for full git clones), then perhaps it'd be good to reduce as much as possible (so yes for reducing automaticity as well then).

I am sort of torn on the same line as you on this: let's maybe wait until our next BIDS maintainers meeting to decide. I think that this issue can defo wait until then.

Remi-Gau:
I removed the large code duplication from `fnirs_automaticity`

Not sure I see anything else to optimize.

@sappelhoff 
What are the next steps? Merge this and then rewrite the history?

sappelhoff:
Thanks!

>  What are the next steps? Merge this and then rewrite the history?

I would say yes!

sappelhoff (approved):

Remi-Gau:
@sappelhoff 
We should plan a session where we are both online to do the history rewrite.

sappelhoff:
agreed, let's DM on mattermost to make this happen. I'm in Berlin timezone wise.

PR #379, opened 2023-06-14
Draft: Face13 updates

Andesha:
This PR is currently a draft, however, the Face13 dataset has been updated to include HED tags!

There will be more commits to updating events files but this is the start.

Other minor fixes include updates to the coordsystem files, etc

cc @christinerogers


effigies:
Marked as draft so it's obvious to reviewers. There's a "Ready for review" button you can press whenever you want to turn it green again.

christinerogers:
this is contributed for **EEGNet** discussions around HED tags, how derivatives are evolving, etc.

eeg_face13/dataset_description.json
```diff
@@ -12,6 +12,9 @@
 	"ReferencesAndLinks": [
 		"http://jov.arvojournals.org/article.aspx?articleid=2121634"
 	],
+	"HEDVersion" : "8.1.0",
 	"DatasetDOI": "10.1167/13.5.22",
```
christinerogers, line 16: Does the Dataset DOI need to be updated with this PR? @Andesha

eeg_face13/dataset_description.json
```diff
@@ -1,5 +1,5 @@
 {
-	"Name": "faceFO",
+	"Name": "Face13-FAC",
```
christinerogers, line 2: hyphens and underscores are reserved in BIDS so we should avoid using here I think.

VisLab:
The events.tsv files still have HED columns that duplicate what is in the JSON file at the top level.  It should be one or the other.

Also, I thought this dataset might have SCORE annotations, but I didn't see any in the example.  Could you point me in the right direction on this?  The tags I saw were all from the standard library, in which case the HEDVersion should just be "8.1.0" or better "8.2.0" since that is the latest.

VisLab:
The validation seems to be failing because of large files.  You will want to cutoff the number of events in the events.tsv file.  If you remove the HED columns in the events.tsv and just keep the top-level JSON sidecar, you may be able to squeak by with full files.

Andesha:
@VisLab just a note: it appears as though the BIDS validator is struggling to retrieve partnered schemas...

image

PR #408, opened 2023-09-02
[MAINT] check format 'IntendedFor' fields in json files

Remi-Gau:
Preferably this should be the non deprecated format: https://bids-specification.readthedocs.io/en/latest/glossary.html#objects.metadata.IntendedFor

So adding a script to check that in CI.

### DIFF
--- a/.github/workflows/check_intended_for.yml
+++ b/.github/workflows/check_intended_for.yml
@@ -1,0 +1,24 @@
+name: check intended for
+
+concurrency:
+    group: ${{ github.workflow }}-${{ github.ref }}
+    cancel-in-progress: true
+
+on: [push, pull_request]
+
+jobs:
+    cehck_intended_for:
+        runs-on: ubuntu-latest
+        steps:
+            -   name: Checkout 🛎️
+                uses: actions/checkout@v3
+            -   uses: actions/setup-python@v4
+                with:
+                    python-version: "3.11"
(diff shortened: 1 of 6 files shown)


effigies:
It would be good to implement bids-uris in some tooling before we start converting examples or even warning.

Remi-Gau:
> It would be good to implement bids-uris in some tooling before we start converting examples or even warning.

you mean like pybids resolving them for example?

In any case fine with me.

I will wrap up this PR and we can leave it hanging until we feel like using it (or not): this was not a lot of work anyway.

Remi-Gau:
ok added a script to maybe help us convert those URI in batches but the diff gets huge because it may alter the formatting of entire json files at times... so may need improvement...

Remi-Gau:
apparently this shows that the new format of URI is not supported for mircroscopy by the current version of the validator: should I open an issue on the validator at least for book keeping?

Remi-Gau:
see https://github.com/bids-standard/pybids/issues/950
see https://github.com/bids-standard/bids-matlab/issues/624

sappelhoff:
> apparently this shows that the new format of URI is not supported for mircroscopy by the current version of the validator: should I open an issue on the validator at least for book keeping?

yes please, we should support BIDS URIs in all our validators, even the legacy one

> It would be good to implement bids-uris in some tooling before we start converting examples or even warning.

agreed

PR #409, opened 2023-09-05
Removed trial annotation from HED FacePerception_events.json example

VisLab:
There is one event file in the HED examples that has two event consecutive event markers at exactly the same time. The `trial` column in the example was annotated as a value column, resulting in duplicate tags when event marker annotations over multiple rows corresponding to the same time are combined.  

The trial information can be correctly incorporated using the curly brace notation, but the hed-javascript validator's handling of this feature is still under development as is the detecting of errors when multiple rows in the `events.tsv` file have the same onset values.

Both types of errors are now detected by the hed-python validator, so we realized the need to correct the examples.

### DIFF
--- a/eeg_ds003645s_hed/task-FacePerception_events.json
+++ b/eeg_ds003645s_hed/task-FacePerception_events.json
@@ -60,8 +60,7 @@
     }
   },
   "trial": {
-    "Description": "Indicates which trial this event belongs to.",
-    "HED": "Experimental-trial/#"
+    "Description": "Indicates which trial this event belongs to."
   },
   "rep_lag": {
     "Description": "How face images before this one was the image was previously presented.",
--- a/eeg_ds003645s_hed_inheritance/task-FacePerception_events.json
+++ b/eeg_ds003645s_hed_inheritance/task-FacePerception_events.json
@@ -43,8 +43,7 @@
     }
   },
   "trial": {
-    "Description": "Indicates which trial this event belongs to.",
-    "HED": "Experimental-trial/#"
+    "Description": "Indicates which trial this event belongs to."
   },
   "stim_file": {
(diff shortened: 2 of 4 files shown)


effigies:
@VisLab I don't feel particularly qualified to verify changes to HED. I'm inclined just to merge, as you know what you're doing, but it also seems unfair not to give your PRs the benefit of a second pair of competent eyes. Are there two people (including yourself if you want) who can approve HED-related pull requests and we can give merge permissions to?

@christinerogers Please feel free to jump in if you or someone in your group would be a good reviewer.

VisLab:
Maybe @dungscout96 would be a good reviewer. 

For the record, maybe I should explain in more detail why this is necessary. HED does not allow repeated tags at the same level. The annotation of the `trial` column appears in the annotation for every line in the `events.tsv` files. There is one line in one of the `events.tsv` files  in this example dataset that has **exactly** the same onset values. 

The Python validator recently implemented detection of duplicate `onset` values and merges the annotations for the lines before validation, because these correspond to the same event marker.  This results in two `Trial/1` tags appearing in the same annotation.  To fix this problem for the moment, I just removed the HED annotation for the `trial` column for this dataset.

Note: The HED can handle this situation using the curly brace notation, but the JavaScript hed-validator that BIDS uses does not yet support curly brace annotation.  (This is in process --- but probably won't be released for several weeks.)  The JavaScript hed-validator also does not yet handle merging annotations with duplicate onsets, so it doesn't detect the problem yet.  In order to have this dataset validate in both the JavaScript and Python validators, I just removed the annotation for the `trial` column.

When the dust settles in a few months and everything has been deployed in both code bases, I'll re-annotate... its a good illustration of the handling of synchronous event markers.

VisLab:
@monique2208 would also be a good reviewer.

monique2208 (approved):

PR #431, opened 2024-03-21
[ENH] add an entry to behavioral datasets to listing of datasets

Remi-Gau:


effigies (approved):
Can't think of any reason why not.

PR #433, opened 2024-04-12
[WIP] examples for BEP044, Stim-BIDS

neuromechanist:
Following https://github.com/bids-standard/bids-specification/issues/153 and https://github.com/bids-standard/bids-specification/issues/751, here is an example of the stimuli-directory with the current suggestions. 

The `sub-sampleTimeVarying` contains an EEG run while the participant was watching a movie.
The `sub-sampleStillImage` [will] contain an iEEG run of a participant while they were watching the Natural Scene Dataset (NSD) images.

We will update the structure as the issues progress.

@dorahermes, is it possible to share an NSD-iEEG data run (or even the file header, or events.tsv)?
@dungscout96, can you add some other annotation files for the present?

- BIDS Specification PR: https://github.com/bids-standard/bids-specification/pull/2022
- BIDS Validator PR: NONE


stimuli_structure/sub-sampleTimeVarying/eeg/sub-sampleTimeVarying_task-ThePresent_eeg.set
Remi-Gau, line 1: Note that large data files have to be truncated:

https://github.com/bids-standard/bids-examples/blob/master/CONTRIBUTING.md#how-to-truncate-data-files-to-0kb

neuromechanist: Thanks for the correction. Truncated the file to 0 kb!

dungscout96:
I added examples of annotations for the movie watching experiment in which two separate sets of annotation files describe two different aspects of the movie (auditory and action viewing). This is in addition to the log luminance ratio annotation already provided by @neuromechanist for the same movie.

dorahermes:
> @dorahermes, is it possible to share an NSD-iEEG data run (or even the file header, or events.tsv)? 

We could share an small template or some empty data (full data to be released in the future), or create an fMRI based example.

stimuli_structure/stimuli/stimuli.tsv
```diff
@@ -0,0 +1,3 @@
+stim_id	type	NSD_id	COCO_id	first_COCO_description	HED
+nsd02951.png	still_image	2951	262145	"an open market full of people and piles of vegetables."	"((Item-count, High), Ingestible-object)), (Background-view, ((Human, Body, Agent-trait/Adult), Outdoors, Furnishing, Natural-feature/Sky, Urban, Man-made-object))"
+the_present_movie.txt	video	n/a	n/a	n/a	n/a
```
bendichter, line 3: shouldn't this be a video filetype, not .txt?

neuromechanist: Very good question.
Here, the movie is under copyright, so, it can't be shared freely with the BIDS dataset. In this case, the text file includes a link to publicly accesbile location for the movie.

neuromechanist: Resolved in the spec, now the stimulus files can be only a json sidecar, especailly when the file is not accesbile like under copyeight.

stimuli_structure/stimuli/stimuli.json
```diff
@@ -0,0 +1,23 @@
+{
+    "stimulus_id": {
+        "Description": "The stimulus ID included in the stimuli directory"
+    },
+    "type": {
+        "Description": "The type of stimulus"
+    },
+    "present": {
```
dorahermes, line 8: Maybe add that this is a boolean? This seems to work as long as it's clarified. The Google Doc there was an optional URL column however.

yarikoptic:
dev-prs validations failing, likely needing also associated bids-validator changes to accommodate changes to bids-schema.  Would require coordination with @effigies and/or @rwblair I guess on that end?

beh_stimuli/stimuli/faces/stimuli.json
```diff
@@ -0,0 +1,9 @@
+{
+    "emotion": {
+        "Description": "Displayed emotion of the face stimulus.",
+        "Levels": {
+            "neutral": "Neutral expression",
+            "happy": "Happy expression"
```
yarikoptic, line 6: shouldn't here be also HED descriptors for each level value similarly to how is done in events.json?

beh_stimuli/stimuli/nsd/stimuli.json
```diff
@@ -0,0 +1,8 @@
+{
+    "description": {
+        "Description": "Natural-language description of the image, generated by a large language model (claude-opus-4-8) from the image content."
+    },
+    "NSD_id": {
+        "Description": "Numeric image identifier in the Natural Scenes Dataset (NSD) shared1000 set. The images derive from the Microsoft COCO dataset and are not redistributed here (present: false); see https://naturalscenesdataset.org for access. HED annotations were generated from the descriptions via the HEDit tool and quality-controlled against the HED schema."
+    }
+}
```
yarikoptic, line 8: I think it would be better to limit this example to just a few (2-3) entries within .tsv (no point really for full list). I am yet to read on `present: False` intended behavior and how user can actually get those images then since I do not see any "URL pattern" or alike to formalize data access to them!

stimuli_structure/stimuli/stim-forrest_part-1.json
```diff
@@ -0,0 +1,5 @@
+{
+	"Type": "audiovideo",
+	"Description": "Segment1, duration 902 seconds"
+
```
yarikoptic, line 4: here  and others worth removing meaningless empty line

stimuli_structure/stimuli/stim-thepresent_annot-auditory_events.tsv
```diff
@@ -0,0 +1,79 @@
+onset	duration	mother_voice	boy_voice	dog_voice	foley_sound	background_music	game_sound	game_voice
+0	226.62
```
yarikoptic, line 2: we have generic approach of `n/a` for .tsv files where value is not present... I wonder if that should be applied here too

stimuli_structure/stimuli/stim-thepresent_annot-LogLumRatio_events.tsv
```diff
@@ -0,0 +1,59 @@
+onset	duration	shot_number	LLR
+0	n/a	video_start	video_start
```
yarikoptic, line 2: what is this row? incongruent with the types

stimuli_structure/stimuli/stim-thepresent_annot-LogLumRatio_events.json
```diff
@@ -0,0 +1,21 @@
...
+        "Units": "seconds",
+        "HED": "(Duration/# s, ({shot_number}))"
+    },
+    "shot_number": {
+        "Description": "Shot sequence number in the movie",
+        "HED": "(Sensory-event, Visual-presentation, (Computer-screen, Label/shotNumber, Item/#))"
+    },
+    "LLR": {
+        "Description": "Log Luminance Ratio for the three frames before and after the shot change",
+        "Units": "dB",
```
yarikoptic, line 18: add [`Format`](https://bids-specification.readthedocs.io/en/stable/glossary.html#objects.metadata.Format) to all of them (here and others where `number` or `integer` or other non-string values?

stimuli_structure/stimuli/stimuli.json
yarikoptic: yeap, as left elsewhere -- feels like need `Format`

stimuli_structure/participants.json
```diff
@@ -0,0 +1,17 @@
+{
+    "participant_id": {
+        "Description": "The prticipant ID set in the HBN dataset"
```
yarikoptic, line 3: typo! don't we have codespell here already? we do... so should trigger


```suggestion
        "Description": "The participant ID set in the HBN dataset"
```

but also is this really for HBN dataset, or should we not disclose here for the sake of example

note that providing this entry, you overload the entire definition of `"participant_id"`, so likely worth adding `"Units" : "years"` to stay close to example in https://bids-specification.readthedocs.io/en/stable/modality-agnostic-files/data-summary-files.html#participants-file .

yarikoptic (changes requested):
initial pass

PR #434, opened 2024-04-13
Fixup and enhance invocation of run_tests.sh

yarikoptic:
ATM it would complain that rc is not legit command due to all those spaces. Because no set -e used it just continues.

Also made it executable and adjusted shebang to be universaly working. There should be no need to invoke directly via bash -- shebang would do that

Script likely worked correctly anyways since there is explicit `exit $rc` so if `$rc` remained undefined and no failed validator runs, it was just `exit` - clean exit?

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ script that is provided in this repository. This script makes use of the
 handles some special case examples (see
 [Validator Exceptions](#validator-exceptions)).
 
-Simply run `bash run_tests.sh` in a command line from the root of the
+Simply run `./run_tests.sh` in a command line from the root of the
 `bids-examples` repository.
 
 ### Validator exceptions
--- a/.github/workflows/validate_datasets.yml
+++ b/.github/workflows/validate_datasets.yml
@@ -86,5 +86,5 @@ jobs:
     - name: Validate all BIDS datasets using bids-validator
       run: |
         cat ./run_tests.sh
-        bash ./run_tests.sh
+        ./run_tests.sh
       shell: bash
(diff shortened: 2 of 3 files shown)


sappelhoff (approved):

yarikoptic:
So the honors would go to @effigies for the last look/click of the green button! ;)

README.md
```diff
@@ -304,4 +304,4 @@ DO NOT EDIT DIRECTLY.
...
-| [synthetic](https://github.com/bids-standard/bids-examples/tree/master/synthetic)                           | A synthetic dataset                                                                         | anat, beh, func                 | T1w, beh, bold, events, physio, scans, sessions, stim                                                        | n/a                                              | [@effigies](https://github.com/effigies)   |
\ No newline at end of file
+| [synthetic](https://github.com/bids-standard/bids-examples/tree/master/synthetic)                           | A synthetic dataset                                                                         | anat, beh, func                 | T1w, beh, bold, events, physio, scans, sessions, stim                                                        | n/a                                              | [@effigies](https://github.com/effigies)   |
```
effigies, line 307: I believe this table is generated automatically. If it will get reverted with the next regeneration, then it would be better to fix up the script.

sappelhoff: probably "no newline" at the end of file? https://github.com/bids-standard/bids-examples/blob/711be5ba05e8071eb9048aa07ce79c30dcecfdbd/tools/print_dataset_listing.py#L129

And it got auto-inserted by the editor the last person touching this file was using?

PR #438, opened 2024-04-20
FIX: IntendedFor errors detected by schema validator

### LINKED ISSUE #437: Bad IntendedFors being caught by schema validator
MEG `coordsystem.json`s have perhaps not been tested up to now.

Should be subject-relative:

https://github.com/bids-standard/bids-examples/blob/f60cd971b5fbeb8ab33f7686e7925a3080eefbf5/ds000117/sub-01/ses-meg/meg/sub-01_ses-meg_coordsystem.json#L31

https://github.com/bids-standard/bids-examples/blob/f60cd971b5fbeb8ab33f7686e7925a3080eefbf5/ds000246/sub-0001/meg/sub-0001_coordsystem.json#L7

Haven't checked them all. These should be fixed upstream in OpenNeuro if possible. Or do we have a way of patching examples until fixed upstream?

effigies:
I am also submitting these fixes to the OpenNeuro dataset authors.

Leaving as draft until I get through all of the validator failures.

Closes #437.

### DIFF
--- a/ds000117/sub-01/ses-meg/meg/sub-01_ses-meg_coordsystem.json
+++ b/ds000117/sub-01/ses-meg/meg/sub-01_ses-meg_coordsystem.json
@@ -28,5 +28,5 @@
  "HeadCoilCoordinateSystemDescription": "5 HPI coils used",
  "AnatomicalLandmarkCoordinateSystemDescription": "For images of anatomical landmarks, see http://imaging.mrc-cbu.cam.ac.uk/meg/SubjectPreparation",
  "FiducialsDescription": "For images of anatomical landmarks, see http://imaging.mrc-cbu.cam.ac.uk/meg/SubjectPreparation",
- "IntendedFor": "sub-01/ses-mri/anat/sub-01_ses-mri_acq-mprage_T1w.nii.gz"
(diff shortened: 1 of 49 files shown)


effigies:
Status

| Dataset | Upstream contacted | Upstream fix released |
|---|---|---|
| ds000117 | Yes | https://openneuro.org/datasets/ds000117/versions/1.0.6 |
| ds000246 | Yes | |
| ds000247 | Yes | |
| eeg_ds003645s_hed_demo | Yes | |
| ieeg_* | No (pending https://github.com/bids-standard/bids-validator/issues/1949) | |

rwblair:
I think the leading slash should be required, it was tripping up the exists/filetree.contains check in the schema validator due to how the string was being split.

https://bids-specification.readthedocs.io/en/stable/common-principles.html#specification-of-paths
"Throughout BIDS all such paths MUST be specified using the slash character (/), regardless of the operating system that a particular dataset is curated on or used on."

effigies:
That's the separator, not the initializer. It's followed with:

> Paths SHOULD NOT be absolute local paths [...]

Which I would interpret to mean that they shouldn't have an initial `/`?

PR #442, opened 2024-04-26
FIX: eeg ds003645s hed demo

effigies:
This fixes two issues with `eeg_ds003645s_hed_demo` identified by the schema validator. 

1) `electrodes.tsv` does not permit a `task` entity.
2) sub-004/ses-1 EEG files lacked a sidecar. I just copied what was consistent in sub-002/003 into a top-level sidecar.


VisLab (approved):
Thanks for fixing.... ds003645 on openNeuro validated and it uses task in the electrodes file names.  I looked in the spec and see that task now isn't in the filename spec for electrodes. Is this new?

effigies:
The validator was written more leniently than the spec. I have recently pushed https://github.com/bids-standard/bids-specification/pull/1722, which re-added the `run` to electrodes to accommodate existing data because it had been mentioned in the spec previously. If it would be useful to re-add task to match existing usage, we could open another PR there as well.

But I would like to go ahead and get this in for the sake of testing anyway. We can re-relax the validator separately, if needed.

VisLab:
@arnodelorme @dungscout96 This may be an issue for bids-matlab-tools.  I have posted [https://github.com/sccn/bids-matlab-tools/issues/212](https://github.com/sccn/bids-matlab-tools/issues/212).

PR #443, opened 2024-05-08
[ENH] add some files to raw for qmri

Remi-Gau:
- relates to https://github.com/bids-standard/bids-specification/pull/1813

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "circleci.persistedProjectSelection": [
+        "gh/Remi-Gau/bids-examples"
+    ]
+}


Remi-Gau:
@agahkarakuzu 

Let me know if there are some obvious files that would make sense to add under raw.

MP2RAGE UNIT1 I have defo seem being generated by the scanner. T1maps not so much.

We could also just say that for the BIDS examples we make sure that every qmri suffix is at least present once under raw to make sure that we get better coverage for the BIDS validator.

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
+    "circleci.persistedProjectSelection": [
+        "gh/Remi-Gau/bids-examples"
+    ]
+}
```
effigies, line 5: Maybe let's add `.vscode` to `.gitignore`? I don't think we want this in the repo.

Remi-Gau: oh dear. 
"git add --all" strikes again

Issue #447, opened 2024-06-06
Convert Range into Levels for 


yarikoptic:
deno based validator in 

- #435 

fails with
```
Validating dataset fnirs_automaticity/: Running  bids-validator fnirs_automaticity --json --ignoreNiftiHeaders
  [
    {
      "key": "TSV_VALUE_INCORRECT_TYPE",
      "severity": "error",
      "reason": "A value in a column did match the acceptable type for that column headers specified format.",
      "files_1": [
        {
          "name": "participants.tsv",
          "path": "/participants.tsv",
          "evidence": "'10' {
    name: \"handedness\",
    display_name: \"Subject handedness\",
    description: 'String value indicating one of \"left\", \"right\", \"ambidextrous\".\n' +
      \"\n\" +
      'For \"left\", use one of these values'... 216 more characters,
    type: \"string\",
    enum: [
      \"left\",         \"l\",
      \"L\",            \"LEFT\",
      \"Left\",         \"right\",
      \"r\",            \"R\",
      \"RIGHT\",        \"Right\",
      \"ambidextrous\", \"a\",
      \"A\",            \"AMBIDEXTROUS\",
      \"Ambidextrous\", \"n/a\"
    ]
  }"
        }
      ]
    }
  ]
```

since values there are numeric 9 and 10:

```
❯ mlr --itsv  cut -f handedness fnirs_automaticity/participants.tsv | sort | uniq -c
     19 handedness=10
      5 handedness=9
```

and .json contains  some ad-hoc `Range` thus trying to change default "Levels" assigned to that column.

```shell
❯ jq .handedness < fnirs_automaticity/participants.json
{
  "Description": "handedness of the participant as a score on the Dutch Handedness Questionnaire",
  "TermURL": "doi:1765/956",
  "Range": "-10 (left) to + 10 (right)"
}
```

I think it is easy to resolve by providing Levels. Alternative is to extend BIDS with support for range, but that is in the future and thus not feasible as a quick fix here.

attn @sappelhoff @rob-luke @robertoostenveld who mastered the original PR as depicted in e84c122a11a4850d9b68ab4a13622e7f9754c26c

effigies:
The issue remains that we do are not yet validating JSON-defined columns.

PR #448, opened 2024-06-07
Remove SKIP_VALIDATION from datasets the deno validator can handle

effigies:
As of https://github.com/bids-standard/bids-validator/pull/1987, these should no longer raise errors.

### DIFF
--- a/.github/workflows/validate_datasets.yml
+++ b/.github/workflows/validate_datasets.yml
@@ -103,12 +103,6 @@ jobs:
         fi
       shell: bash
 
-    - name: Mark to be skipped some examples for a deno based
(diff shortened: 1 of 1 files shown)


yarikoptic:
Wow -- it is green, great!

Would it still be feasible to request some kind of an option/switch to cause them to become ERRORs so we could detect/fix them up, such as 

- #447 

in then examples? (I don't know if it would be feasible to aim to fix all the WARNINGs since didn't check how many we do have ATM)

yarikoptic (approved):

effigies:
Yes, I think we need a CLI or config option to override the severity level of issues. If that's not already an issue on the validator, we should add it.

PR #449, opened 2024-06-11
Use deno compile to install deno bids-validator

yarikoptic:
Make it less cumbersome and may be more efficient since produces a binary executable file

Based on info from https://github.com/bids-standard/bids-validator/pull/1942#issuecomment-2159539424

### DIFF
--- a/.github/workflows/validate_datasets.yml
+++ b/.github/workflows/validate_datasets.yml
@@ -71,11 +71,9 @@ jobs:
       run: |
         LOCAL_BIN=$HOME/.local/bin
         VALIDATOR=$LOCAL_BIN/bids-validator
-        mkdir -p $LOCAL_BIN
-        export PATH="$LOCAL_BIN:$PATH"
-        echo PATH="$PATH" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


yarikoptic:
doesn't work on windows
```
Download https://dl.deno.land/release/v1.44.1/denort-x86_64-pc-windows-msvc.zip
/d/a/bids-examples/bids-examples
D:\a\_temp\b1eee1f8-8f1c-4dae-9944-289f42dbd60a.sh: line 8: bids-validator: command not found
Error: Process completed with exit code 127.
```
so screw it for now

.github/workflows/validate_datasets.yml
```diff
@@ -71,11 +71,11 @@ jobs:
       run: |
         LOCAL_BIN=$HOME/.local/bin
         VALIDATOR=$LOCAL_BIN/bids-validator
-        mkdir -p $LOCAL_BIN
-        export PATH="$LOCAL_BIN:$PATH"
-        echo PATH="$PATH" >> $GITHUB_ENV
```
effigies, line 76: Removing `$HOME/.local/bin` from the path seems the likely cause of the failure to find the executable.

PR #454, opened 2024-07-11
fix(7t_trt): Define handedness column to match numeric values

effigies:
participants.tsv contains numeric values. The [Edinburgh Handedness Inventory](https://www.physio-pedia.com/Edinburgh_Handedness_Inventory_(EHI)) seems a likely source for these values. In any event, they do not match the BIDS default definition of "right", "left" and "ambidextrous".


yarikoptic (approved):
Looks good to me

Issue #459, opened 2024-07-31
MTS example do not follow current specs


bpinsard:
In https://osf.io/k4bs5/ `ds-mtsat` uses outdated/pre-BEPS `_acq-<MTOn|MTOff|T1w>`  rather thank `_flip-<index>_mt-<on|off>` which can confuse willing BIDSonauts. ;)

Remi-Gau:
So in general we do allow some divergence between examples and the data that were used to generate them. But indeed this can be confusing...

@agahkarakuzu 
would you be OK if we moved those examples to openneuro?

This actually may be a good policy for examples in general because would help make sure that the datasets are at least kept valid (and easier to update)

bpinsard:
Another thing I noticed in the examples are the very scarce metadata.
If the dicoms still exists, it would be great to have all the metadata that dcm2niix now produces by default in the sidecar.
I am mainly thinking about pulse sequences info and other MR related options.

bpinsard:
Also regarding continuous BIDS-compliance testing, if the dataset is datalad with a fork hosted on github for the git part (full dataset can be on openneuro), we can set up github actions to run the bids-validator when changes are made and also scheduling it (~1/month) with latest bids-validator. 
for example:
https://github.com/courtois-neuromod/anat/blob/main/.github/workflows/bids-validator.yml

PR #461, opened 2024-08-06
fix: (Re)move sidecars with no data files

effigies:
Detected by https://github.com/bids-standard/bids-validator/pull/2057.

@Remi-Gau I think you're clearer on the processes here. Is there a process for fixing upstream?


Remi-Gau:
> Is there a process for fixing upstream?

not really: several examples have started to diverge from their "source" data and AFAICR we always said this was fine, though as mentioned [here](https://github.com/bids-standard/bids-examples/issues/459#issuecomment-2264838750) it would be nice if the target data was on openneuro and "we" (bids maintainers and dataset curators) were given access to it.

effigies:
Going to go ahead and merge, as these are invalid files.

PR #467, opened 2024-09-17
HED-SCORE example of EEG and iEEG data with annotations 

dorahermes:
This BIDS example has EEG and iEEG data with annotations from the HED-SCORE library schema. It includes a subject with EEG data where artifacts are annotated, a subject with EEG data where seizures are annotated and a subject with iEEG data during photic stimulation was done and this is annotated as a modulator. Examples are provided of artifacts or seizures only observed in certain channels using the channel column in the _events.tsv file.

This example uses the recently released [HED-SCORE version 2.0.0](https://github.com/hed-standard/hed-schemas?tab=readme-ov-file) and replaces pull request #324 . 

It passes the BIDS online validator, comment welcome!

@VisLab @christinerogers


christinerogers:
Hi @dorahermes I'm noticing bids-validator errors in the checks.  Is this something expected / unexpected / easily solvable?

dorahermes:
Thank you so much for checking! I am not sure I can reproduce the error, as I did not get any errors using the online validator, except for the fact that the .edf files are empty, which I believe is required for the BIDS examples.

image

christinerogers:
Reporting back from the Maintainers meeting -- A fix is on the way from the maintainers group since this is impacting lots of datasets, but in the meantime per @effigies :   
Since this is passing on a few branches including`master-deno` and 'master' this should be good to merge.  
I can approve on that basis @dorahermes

christinerogers (commented):
approving

PR #470, opened 2024-11-05
CI: Test stable (deno), dev (deno) and legacy (node)

effigies:
We were previously testing against the stable and dev versions of the NodeJS validator and only the dev version of the Deno validator.

This now tests against the stable and dev versions of the Deno validator and only the stable version of the NodeJS validator, reflecting the expectation that further development will be almost exclusively in the Deno validator.


effigies:
@Remi-Gau @yarikoptic Any comments? I'll merge soon just to get CI working again, but this can be reconfigured if you think we should do something differently.

Remi-Gau:
Good with me from a quick read

yarikoptic:
thank you!

PR #476, opened 2024-12-19
Enhance 7t_trt example with demonstration of "entities-less" physio.json for inheritance principle

yarikoptic:
We already have a single dataset with similar entities less file: ds000248/T1w.json . I decided to enhance this one as it is a nice demonstration of allowance here since there is another metadata file but with a different suffix

To complement
- https://github.com/bids-standard/bids-specification/pull/2019


effigies (approved):

yarikoptic:
Let's seek few more opinions just to be sure ;)  requesting from experts on inheritance and those who touched this example in the past

Remi-Gau (approved):

PR #477, opened 2024-12-19
Make possible to run on specified datasets, address some shellcheck concerns (quotes)

yarikoptic:
There are some remaining issues for shellcheck but decided not to touch here.


yarikoptic:
seems like `'...'` quotes I added had negative effect! will remove

PR #479, opened 2025-01-17
BIDS-Prov demo using `dcm2niix` on hirni-demo data

bclenet:
This PR proposes to add an example dataset `hirni-demo`, based on https://github.com/psychoinformatics-de/hirni-demo, converting the source data from DICOM to Nifti + json sidecars using `dcm2niix`.

It is an opportunity to add provenance records (see `hirni-demo/prov/hirni-demo.prov.jsonld`) as specified by the BEP028: BIDS-Prov.


satra:
@bclenet - we should make sure the prov-jsonld documents go through a json-ld parser/validator.

PR #480, opened 2025-01-30
[WIP] BEP042 - Emg examples

neuromechanist:
The emg-example dataset contains the examples discussed during the biweekly meetings of the BEP042 team @bids-standard/bep042. Each subject is one of the examples from the [metadata document](https://docs.google.com/document/d/1s4SbStsm2VXqFbHEVF6Og7BCtjsRUDhzcZj9LO07r6Q/edit?tab=t.0).

Here are some key metadata discussed which might be different from other modalities:
1. `channels.tsv` optionally includes _placement_ and _group_ columns borrowed from Motion and iEEG, respectively.
2. `channels.tsv` also optionally includes _signal_electrode_ column, indicating which electrode(s) was used (in conjunction with the _reference_) to make the channel.
3. Multiple coordinate systems can be added to the `channels.tsv` and `electrodes.tsv` to adequately describe the channel or electrode placement.
4. The coordinate systems can be optionally linked together using _Anchor_ electrodes.
5. A _loc_reference_ column is optionally added to indicate which coordinate system is used for _[x,y,z]_ columns in `electrodes.tsv` or the _placement_ column in `channels.tsv`.

We appreciate your input and feedback on the examples. You can comment about specific examples here, on the original issue https://github.com/bids-standard/bids-specification/issues/1371, or if needed the PR https://github.com/bids-standard/bids-specification/pull/1998

> [!Note]
> **We meet regularly to discuss this BEP**
> 
> Next bi-weekly meeting: 12 February 2025 on https://ucsd.zoom.us/j/96433382377
> 
> Communication channel on github repo and matrix: https://github.com/bids-standard/bids-specification/issues/1371


neuromechanist:
Thanks much @effigies,
I resolved some issues with the examples.
Remaining is the `space-<label>` is not yet allowed for `coordsystem.json` and the values under `coordinate_system` column under `electrodes.tsv` are not associated with the `<label>` of the `space-<label>`. I think @drammock wanted to add these to https://github.com/bids-standard/bids-specification/pull/1998.

effigies:
Sorry, I thought I'd pushed my changes to the schema. I've merged with @drammock's recent changes and pushed.

effigies:
Remaining issues:

* emg_CustomBipolarFace:

```
 	[ERROR] EMG_COORD_SYS_MISMATCH Some values in the coordinate_system column of *_electrodes.tsv are not present as
space entities in *_coordsystem.json files.

		/sub-01/emg/sub-01_electrodes.tsv
```

* emg_TwoWristbands:
```
	[ERROR] TSV_ADDITIONAL_COLUMNS_MUST_DEFINE Additional TSV columns must be defined in the associated JSON sidecar for this file type
		placement
		/sub-01/emg/sub-01_task-typing_channels.tsv

	Please visit https://neurostars.org/search?q=TSV_ADDITIONAL_COLUMNS_MUST_DEFINE for existing conversations about this issue.

	[ERROR] JSON_KEY_REQUIRED A JSON flle is missing a key listed as required.
		EMGCoordinateUnits
		/sub-01/emg/sub-01_space-leftForearm_coordsystem.json - Field description: Units of the coordinates of `EMGCoordinateSystem`.

		/sub-01/emg/sub-01_space-rightForearm_coordsystem.json - Field description: Units of the coordinates of `EMGCoordinateSystem`.


	Please visit https://neurostars.org/search?q=JSON_KEY_REQUIRED for existing conversations about this issue.
```

neuromechanist:
> emg_CustomBipolarFace:
``` 	
[ERROR] EMG_COORD_SYS_MISMATCH Some values in the coordinate_system column of *_electrodes.tsv are not present as
space entities in *_coordsystem.json files.
		/sub-01/emg/sub-01_electrodes.tsv
```


Down to one.
This is interesting. The `coordsystem.json` does not have `space-<label>` since there is only one file there. `electrodes.tsv` also does not have a `coordinate_system` column. I think this should validate. WDYT @drammock?

drammock:
> I think this should validate. WDYT @drammock?

I agree that it seems valid. I cannot reproduce this failure locally:

```console
$ cd ../validator/
$ git log --oneline -1
79248e84 (HEAD -> feat/add-keys-to-assoc-coordsys, origin/feat/add-keys-to-assoc-coordsys) changelog
$ cd ../examples/
$ git log --oneline -1
14abe5f8 (HEAD -> emg_examples, neuromechanist/emg_examples) fixes to channels.tsv and electrodes.tsv in some datasets? (#3)
$ cd ../spec/
$ git log --oneline -1
a55ff9ef (HEAD -> emg, origin/emg) schema: Check for coordsystem or coordsystems
$ uv run bst export > src/schema.json
$ DATASET=emg_CustomBipolarFace /opt/bids/validator/local-run --schema file:///opt/bids/spec/src/schema.json /opt/bids/examples/$DATASET
        [WARNING] SIDECAR_KEY_RECOMMENDED A data file's JSON sidecar is missing a key listed as recommended.
                StimulusPresentation
                /sub-01/emg/sub-01_task-jumping_events.tsv - Field description: Object containing key-value pairs related to the software used to present
the stimuli during the experiment.


                EpochLength
                /sub-01/emg/sub-01_task-jumping_recording-highDensity_emg.edf - Field description: Duration of individual epochs in seconds (for example, `1`)
in case of epoched data.
If recording was continuous or discontinuous, leave out the field.

                /sub-01/emg/sub-01_task-jumping_recording-bipolar_emg.edf - Field description: Duration of individual epochs in seconds (for example, `1`)
in case of epoched data.
If recording was continuous or discontinuous, leave out the field.


        Please visit https://neurostars.org/search?q=SIDECAR_KEY_RECOMMENDED for existing conversations about this issue.


          Summary:                         Available Tasks:        Available Modalities:
          18 Files, 132 kB                 jumping                 emg                  
          1 - Subjects 1 - Sessions                                                     

        If you have any questions, please post on https://neurostars.org/tags/bids.
```

effigies:
Pushed https://github.com/bids-standard/bids-specification/pull/1998/commits/b1fea70a4484832269c46c3b576e65cfbb9d8d60, which resolved that issue. LMK if you disagree with the logic and think it needs a different solution.

The only other one I can think of is requiring `coordsystem.json` to have an identifier (e.g., `EMGCoordinateSystemName`). Then we would switch to checking that `ParentCoordinateSystem` matches against those values, and require `space-<label>` (when present) would have the same value. That is not dramatically more complicated than what is currently there, from a validator perspective.

I also want to draw your attention to https://github.com/bids-standard/bids-validator/pull/268/files#r2365617071:

> The EMG examples do not have this case, but with multiple coordsystem.jsons, you could imagine having some at the root and some at the leaf.
> 
> We currently strongly assume [...] that the first [directory] level found is the level of the association, which has worked for single-file associations. This probably needs to be relaxed, but the logic will be more intricate. I think you would need to take the first coordsystem.json with a given space entity, not the full collection.
> 
> For now this could be noted as a limitation in the validator implementation.

Does this limitation seem acceptable in the short term?

drammock:
> Pushed [bids-standard/bids-specification@b1fea70](https://github.com/bids-standard/bids-specification/commit/b1fea70a4484832269c46c3b576e65cfbb9d8d60), which resolved that issue. LMK if you disagree with the logic and think it needs a different solution.

That solution looks good to me. 

> [...] Does this limitation seem acceptable in the short term?

Is the limitation that you can't collate coordsystem.json files at different directory levels? Or is it that multiple coordsystem.json files with the same space entity can't be collated? The latter seems fine;
I doubt it will come up. The former should be fine in the short term but should be fixed eventually. 

@neuromechanist do you agree?

effigies:
> Is the limitation that you can't collate coordsystem.json files at different directory levels? Or is it that multiple coordsystem.json files with the same space entity can't be collated?

The first. Or at least, we don't currently look past the first directory to have responsive files. In the case of sidecars, we get all responsive files, because we need to aggregate their contents. Associations have been just "find the first file".

I suppose an interesting question here is: Should `/space-A_coordsystem.json` and `/sub-01/emg/sub-01_space-A_coordsystem.json` have their contents aggregated like a sidecar, or should `/sub-01/emg/sub-01_space-A_coordsystem.json` take precedence?

neuromechanist:
I believe that most of the `coordsystem.json` for EMG should/will live under the dataset root, not at subject level as they don't carry subject/session specific keys by default. This is different from MEG and EEG, where specific fiducial points can be optimally defined in `coordsystem.json`. In those cases,  most examples I saw have this file under session or modality.

So, I think we might need, at least for one example if not most, to move the `coordsystem.json` files to the root to account for this case.

> I suppose an interesting question here is: Should /space-A_coordsystem.json and /sub-01/emg/sub-01_space-A_coordsystem.json have their contents aggregated like a sidecar, or should /sub-01/emg/sub-01_space-A_coordsystem.json take precedence?

Indeed, this is interesting, as this is not formally a sidecar, but hopefully should be treated as a sidecar. IMO, `coorsystem.json` for ephys is a glorified `electrodes.json`. It would be great if we merge it into `electrodes.json`  in BIDS2.0, in favor of having one fewer file 😁.

neuromechanist:
I moved the `coordsystem.json` for the TwoWristbands example to root to make cover this use case as well.

effigies:
@neuromechanist BIDS examples should not include full data files, but should be truncated to zero bytes. If header data is needed for some validation rules, then one dataset could have file headers.

If you have real data to use for examples, the best thing would be to host it elsewhere, and have a script for updating the example as any changes are made. The [dataset_listing.tsv](https://github.com/bids-standard/bids-examples/blob/master/dataset_listing.tsv) allows links to the real data to be shared.

(I do see that we have not consistently enforced this, and as a result this repository is now 162MB on a fresh clone.)

drammock:
> BIDS examples should not include full data files, but should be truncated to zero bytes

That is probably my fault, I filled some files with 1kb random data to make the validator warning go away. @neuromechanist can you fix? I'm away from the office this week.

neuromechanist:
Well, I changed the original 1kB files to be a little more (around 100kB on average), to have actual EDF file headers and half a second worth of "simulated" data respecting the channel count and sampling frequency.

This has a benefit of the examples being standalone and work properly with tools, rather than linking to adhoc cloud storage.

Happy to move them outside. LMK.

PR #481, opened 2025-02-05
Improve synthetic with + in task label for `beh/` data

yarikoptic:
For
- https://github.com/bids-standard/bids-specification/pull/1926

mimicing a very hypothetical case where e.g. background color in stroop test is different


tsalo:
The example is failing in the expected way, which is great!

Remi-Gau (approved):

effigies:
This will render the synthetic dataset unvalidatable until a new release of BIDS. I think a small example would make more sense, and then it could be limited to the `dev` tests.

PR #497, opened 2025-05-28
 [BEP028] provenance - SPM preprocessing

bclenet:
This example aims at showing provenance records for a fMRI preprocessing, performed by [SPM](https://www.fil.ion.ucl.ac.uk/spm/).

The PR also adds a new table to the README.md for datasets related to provenance (cf. PR https://github.com/bids-standard/bids-examples/pull/495 )


bclenet:
Hi @effigies @rwblair,

With this example, we want to show both:

- the output files of a SPM pipeline (with names as written by SPM);
- the provenance metadata that SPM could generate to be compliant with the BEP.

At the same time, we want to BIDS validate all the provenance metadata. Some of this metadata is stored in sidecars... that will always fail to be validated considering their names are not compliant.

I can't see any good solution for that based on a `.bidsignore` file or having a `derivatives/` folder.

Could you please let me know if you have an idea about that.

Thanks !

PR #503, opened 2025-06-28
Clarified in the README where sourcedata can be found 

dorahermes:
The README now more clearly states where the sourcedata can be found, added a link to OpenNeuro.


VisLab (approved):
LGTM

PR #509, opened 2025-09-02
fix: Update column definitions that do not match column contents

effigies:
Working on improving column validation in the validator, we decided that failure to match declared column values should be an error, not a warning (https://github.com/bids-standard/bids-validator/pull/247). This produces errors in a couple examples, so this PR fixes those examples.

@arnodelorme You are listed as the maintainer of eeg_rishikesh
@markmikkelsen You are listed as the maintainer of mrs_fmrs

Would you mind reviewing these changes?


markmikkelsen (approved):
Changes look good to me.

PR #513, opened 2025-09-11
fix: Align column values and definitions for custom columns

effigies:
The validator has previously only checked that custom columns are defined, and not validated their contents. bids-standard/bids-validator#254 aims to change that. This PR fixes errors that were detected in the examples.


effigies:
In the absence of objections, merging. Post-merge reviews are appreciated!

PR #515, opened 2025-09-18
ENH: add ability to test examples PR against PR in bids-specification

yarikoptic:
Many PRs are developed to accompany WiP BEP PRs. It is beneficial to establish testing against likely a modified BIDS schema in those PRs.  In this solution we should expand matrix of runs with `bids-pr` run against that arbitrary PR if entered in a line with `bids-specification-pr:`.

TODOs
- test with proposed line in the template
- Needs testing against a bids-specification-pr: NUMBER-or-URL
  - [ ] should not run/fail if not a match
  - [ ] should run correctly if number
  - [ ] should run correctly if URL
  - [ ] should rerun when description edited

### DIFF
--- a/.github/workflows/validate_datasets.yml
+++ b/.github/workflows/validate_datasets.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches: ['**']
   pull_request:
+    types: [opened, synchronize, reopened, edited]
     branches: ['**']
   create:
     branches: [master]
--- a/.github/workflows/validate_datasets.yml
+++ b/.github/workflows/validate_datasets.yml
@@ -20,16 +20,47 @@ defaults:
     shell: bash
 
 jobs:
+  # Prepare the matrix dynamically based on whether a bids-specification PR is referenced
+  prepare-matrix:
+    runs-on: ubuntu-latest
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+      bids_pr: ${{ steps.find-pr.outputs.pr_number }}
+    steps:
+    - uses: actions/checkout@v5
+      with:
+        fetch-depth: 0  # Get full history for commit message scanning
+
+    - name: Check for bids-specification PR reference
+      id: find-pr
+      run: |
+        # Also check PR description if this is a PR
+        if [ "${{ github.event_name }}" == "pull_request" ]; then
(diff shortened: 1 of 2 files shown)


effigies:
Please see #480 and #514 for examples of testing BEPs against schema and validator branches.

yarikoptic:
Yep, there temporarily changes the workflow. This is what this tries to avoid and make it more explicit

.github/workflows/validate_datasets.yml
```diff
@@ -65,7 +100,7 @@ jobs:
           deno install -Agf -n bids-validator jsr:@bids/validator
 
     - name: Install BIDS validator (main)
-      if: matrix.bids-validator == 'main'
+      if: matrix.bids-validator == 'main' || matrix.bids-validator == 'bids-pr'
```
effigies, line 103: In general, `dev` tracks unreleased changes to the schema, while `main` tracks the latest released version of BIDS. PRs against the spec will mostly aim for `dev`.

effigies (commented):
The current process is to prevent non-`dev` runs from testing new examples, and use validator and schema from releated PRs in `dev`. Instead of adding a new `bids-pr` run, I think it would be helpful just to automate the changes to `dev`.

If you want the run name to be clearly different from `dev`, then I would replace `dev` with the other name in the matrix, but have the new name use identical steps to `dev`.

As to triggers, I don't really like the false positive rate of these triggers. I wonder if we could use a `label` trigger. Maybe the job could remove the trigger, but we could at least remove and re-add the label to re-trigger.

WDYT?

yarikoptic:
> If you want the run name to be clearly different from `dev`, then I would replace `dev` with the other name in the matrix, but have the new name use identical steps to `dev`.

recommendations? could be `master` or `main` I guess to reflect that it is a branch

.github/workflows/validate_datasets.yml
yarikoptic: yes, but it is useful to know if it fails against released version since if it doesn't but expects to (e.g. new file type was introduced like `atlas-<label>_description.json` but released validator+schema is ok with that means that either they miss something or something else is not stringent enough.  The same about `dev` as -- have we made development validator and/or schema less stringent, as we would expect then only the `dev-pr` to pass in such cases.   That is why I preferred to keep it separate and additional matrix run.

PR #519, opened 2025-11-04
 [BEP028] provenance - study dataset + manual annotation

bclenet:
This example aims at showing provenance for a study dataset.

The usecase is manual brain segmentations performed by several experts from a single T1w file.


bclenet:
@cmaumet would you agree on completely removing the software provenance objects in this example ?
I'm afraid it creates an ambiguity as with the current version we are not able to describe if the software was solely responsible for generating the data or if a human intervened in the process.

PR #523, opened 2025-12-15
[BEP047] Add behavioral audio/video recordings example

bendichter:
This PR adds a new example dataset: `beh_audio_video_recordings/`.

It demonstrates the proposed audio/video behavioral recording support (see bids-standard/bids-specification [PR #2231](https://github.com/bids-standard/bids-specification/pull/2231)) including:

- `_audio.*` and `_video.*` recordings in `sub-*/beh/`
- corresponding sidecar JSON files
- multi-camera recordings using the `recording` entity
- audio runs using the `run` entity
- `_events.tsv`/`_events.json` annotations for speech/vocalization tasks
- `*_scans.tsv` for acquisition timing alignment

Note: media files are small placeholder files for example purposes.

- BIDS Specification PR: https://github.com/bids-standard/bids-specification/pull/2231
- BIDS Validator PR: None


neuromechanist:
Overall LGTM.
Small point, that might be more actually on the spec side, is how accretion we want to annotate keys such as `CameraPosition`? I imagine allowing HED in these sidecars would resolve the ambiguity to a degree.

-----
Will do a formal review one out of draft 😉

yarikoptic:
@bendichter now that we have a reviewable draft of that BEP, I have adjusted PR description to point to that PR so that validation uses that schema.  Didn't wait/check but I would expect that it needs tune up to pass.  I will also backref this PR from the BEP PR.

PR #552, opened 2026-04-15
add sv2a pet atlas from NRU to examples

bendhouseart:
Adds PET atlas to examples, will fail validator for version v1.11.1, fix coming in minor release v1.11.2.


effigies:
You're going to add hemi for volumetric data? Or are you using the FS-NIfTI hack for surfaces?

Either way, just use GIFTI.

CPernet:
I am quite certain the extension should be .mimap and not pet since the values are not PET values anymore (discussed many times in many different threads @mnoergaard

effigies:
If we want to use mimap, we need to finish off https://github.com/bids-standard/bids-specification/pull/2262.

CPernet:
I agree - we cannot push a PET atlas with the wrong extension

bendhouseart:
> I agree - we cannot push a PET atlas with the wrong extension

We need to update the spec then as [any suffix is allowed](https://github.com/bids-standard/bids-specification/blob/6764039d06a93d2bfa203405477f389323814909/src/derivatives/atlas.md?plain=1#L50). I'd rather change the schema than try to meddle with the spec directly, but maybe that's just me. [This PR](https://github.com/openneuropet/bids-specification/pull/1) works with these examples and complies with the spec, so it is possible to have both.

bendhouseart:
Inherit some PET metadata:

- [ ] Tracer Name
- [ ] Tracer Radionuclide
- [ ] tbd

melanieganz:
@bendhouseart, while in principle this is ok, but there is a mistake wrt the endings. It should be pet.gii.gz for files on the surface, aka in the tpl-fsaverage folder, and nii.gz in the tpl-MNI152 folders. This is currently switched around, which doesn't make sense. Maybe we should add that to the validator somehow as well to avoid surface files being added into volume templates? But maybe that is out of scope.

bendhouseart:
> @bendhouseart, while in principle this is ok, but there is a mistake wrt the endings. It should be pet.gii.gz for files on the surface, aka in the tpl-fsaverage folder, and nii.gz in the tpl-MNI152 folders. This is currently switched around, which doesn't make sense. Maybe we should add that to the validator somehow as well to avoid surface files being added into volume templates? But maybe that is out of scope.

Done with [d82051d](https://github.com/bids-standard/bids-examples/pull/552/commits/d82051d0f9d3d3a6b76c9ce4b4a0081cbdc7a8e5)

Per Chris' suggestion (to comply with gifti rules) also updating with `.func.gii` in commit #054c02edde

PR #563, opened 2026-09-09
Add duration column to _scans.tsv

yarikoptic:
To accompany
- https://github.com/bids-standard/bids-specification/pull/2508

Add the optional `duration` column (wallclock seconds from first to end of last acquired sample) to three existing datasets and one new scans.tsv:

- synthetic (MRI, 5 sub × 2 ses): T1w=240s, nback-run-01/02=720s (=288×TR), rest=600s (=240×TR).  All durations are multiples of TR=2.5s and fit within the acq_time gap to the next scan (no acquisition overlaps).

- ds000117 (MEG+EEG): duration=450s (=7.5 min) for all 6 runs per subject (16 subjects), consistent with the TaskDescription in the sidecar JSON ("six, 7.5 minute runs").  Verified no overlap: minimum start-to-start gap across all 16 subjects is 556s (sub-16 run-01→run-02), leaving a 106s residual.

- ds000246 (MEG, CTF): duration=360s for both AEF runs (matching RecordingDuration=360 in each run sidecar); duration=30s for the emptyroom noise run (RecordingDuration=30).

- ds000248 (MEG, Elekta/single-subject): duration=277.71s for the audiovisual run (RecordingDuration≈277.71 in sidecar); duration=110.0s for the emptyroom (RecordingDuration≈109.999).

- volume_timing (fMRI timing variants, new scans.tsv): durations derived from VolumeTiming + per-frame duration metadata in each sidecar.  The acq-deprecated run (VolumeTiming 0–14s, AcquisitionDuration=1.0s per frame) gets duration=15.0s, consistent with the spec rule that the BOLD+VolumeTiming exclusion applies (AcquisitionDuration was used there as a per-frame measure, so the ScansAcquisitionDurationConsistency check is correctly not triggered).

- qmri_qsm (MRI with _part-, new scans.tsv): both part-mag and part-phase entries carry identical acq_time and duration (480s), illustrating that split-storage files of the same acquisition share temporal metadata.

- BIDS Specification PR: https://github.com/bids-standard/bids-specification/pull/2508
- BIDS Validator PR: https://github.com/bids-standard/bids-validator/pull/449
- TODOs:
  - [ ] Drop TEMP commit after verying that we trigger CI to fail with failing data

Co-Authored-By: Claude Code 2.1.263 / Claude Sonnet 4.6 <noreply@anthropic.com>


volume_timing/sub-01/sub-01_scans.tsv
```diff
@@ -0,0 +1,7 @@
+filename	duration
+func/sub-01_task-rest_acq-clusteredST_bold.nii.gz	15.0
+func/sub-01_task-rest_acq-clusteredTA_bold.nii.gz	15.0
+func/sub-01_task-rest_acq-constantDelay_bold.nii.gz	17.0
```
yarikoptic: let's brake it!!!


```suggestion
func/sub-01_task-rest_acq-clusteredTA_bold.nii.gz	111.0
func/sub-01_task-rest_acq-constantDelay_bold.nii.gz	170.0
```

yarikoptic:
hm , TEMP is there and validator is still happy -- likely need to tune up the "breakage" example

effigies:
Your TEMP commit changes the duration for a file with `VolumeTiming`, which is specifically excluded from your schema check. Your schema check requires the file to have `AcquisitionDuration` defined, `VolumeTiming` undefined, and not be BOLD or ASL.

PR #541, opened 2026-02-12
Add SWI example dataset

astewartau:
## Summary

Add example dataset demonstrating susceptibility weighted imaging (SWI) organization in BIDS:

- **Raw data**: 2-echo MEGRE with magnitude and phase components
- **Derivatives**: `swi`, `minIP`, and `Chimap` outputs with `Sources` metadata linking back to raw inputs
- Entry added to `dataset_listing.tsv`

Companion to https://github.com/bids-standard/bids-specification/pull/2343, which adds the `swi` and `minIP` suffixes to the specification.

Ref: https://github.com/bids-standard/bids-specification/issues/1566#issuecomment-3878452407

## Test plan

- [ ] Example validates against the schema once the specification PR is merged

PR #565, opened 2026-09-18
[EXAMPLE] Add raw and derivative SWI example datasets

fabriciocravo:
Cherry-picked from @astewartau's commit (#541).

This adds two SWI dataset examples: one with SWI images in derivatives, another with SWI images in raw data. This split is necessary since SWI can either be generated directly by the scanner or reconstructed as a derivative from raw phase and magnitude images. As evidenced by Issue bids-standard/bids-specification#2489.

Also removes the minIP images and Chimap from the derivatives example, to keep it minimal and scoped to what's currently under review. minIP is still unresolved in bids-standard/bids-specification#2343, so it's better served by its own example once that settles. It also clears up the scope confusion around the original PR.

Ref: bids-standard/bids-specification#2343, bids-standard/bids-specification#2538, bids-standard/bids-specification#2489

@effigies  — I think this example is still incomplete. I should probably add the mask, or other raw files the scanner uses to reconstruct SWI. That said, the mask is processed from phase data rather than raw data off the scanner, so I'm not sure it belongs here. I am wondering if it would be in the scope of this example. Also wondering if the JSON sidecars should include scanner metadata.

@astewartau — what do you think of these changes?

Portions of this PR description were edited with LLM assistance (Claude) for clarity and flow.


effigies:
>  I think this example is still incomplete. I should probably add the mask, or other raw files the scanner uses to reconstruct SWI.

That's not necessary. BIDS isn't about harvesting everything a scanner generates, but about storing useful products. If this is paired with a PR that only adds SWI, the only new thing it should add is SWI.

> That said, the mask is processed from phase data rather than raw data off the scanner, so I'm not sure it belongs here. I am wondering if it would be in the scope of this example.

I think not.

> Also wondering if the JSON sidecars should include scanner metadata.

JSON sidecars should include (a) required metadata; (b) other metadata that would be helpful for a reader to see. I updated with (a).

